PR #6, opened 2017-08-17
Ensure output directory exists when set

### LINKED ISSUE #5: README configuration causes error on first run of stackprof
When running `TEST_STACK_PROF=1 rspec` with the config mentioned in the README

```
TestProf.configure do |config|
  # the directory to put artifacts (reports) in ("tmp" by default)
  config.output_dir = "tmp/test_prof"

  # use unique filenames for reports (by simply appending current timestamp)
  config.timestamps = true

  # color output
  config.color = true
end
```

Produces an error if "tmp/test_prof" doesn't exist. Ideally, TestProf would create this directory if it doesn't exist.

danielwestendorf:
Fixes #5


lib/test_prof.rb
```diff
@@ -91,10 +92,15 @@ class Configuration
     def initialize
       @output = $stdout
       @color = true
-      @output_dir = "tmp"
+      @output_dir = "tmp/test_prof"
```
palkan: Shouldn't we ensure that that dir exists?

I think, we can move `mkdir_p` to `TestProf.artefact_path` method to create directory only when necessary.

lib/test_prof.rb
```diff
@@ -59,8 +59,10 @@ def asset_path(filename)
       ::File.expand_path(filename, ::File.join(::File.dirname(__FILE__), "..", "assets"))
     end
 
-    # Return a path to store artefact
-    def artefact_path(filename)
+    # Return a path to store artifact
+    def artifact_path(filename)
```
palkan: 😄 Actually, it's not a typo ([see](https://en.oxforddictionaries.com/definition/artefact)).

But I'm ok with the change, US variant is more popular.

danielwestendorf: TIL!

lib/test_prof.rb
danielwestendorf: This change should be done.

palkan:
Thanks! Merged and released as v0.2.0.

PR #17, opened 2017-08-31
Combine RSpecStamp with EventProf and RSpecDissect

palkan:
Make it possible to automatically mark _slow_ examples and groups with custom tags. For example:

```sh
EVENT_PROF="sql.active_record" EVENT_PROF_STAMP="slow:sql" rspec ...
```

After running the command above the top 5 slowest example groups would be marked with `slow: :sql` tag.

### DIFF
--- a/lib/test_prof/recipes/rspec/let_it_be.rb
+++ b/lib/test_prof/recipes/rspec/let_it_be.rb
@@ -22,7 +22,7 @@ module TestProf
     # Use uniq prefix for instance variables to avoid collisions
     # We want to use the power of Ruby's unicode support)
     # And we love cats!)
-    PREFIX = "@😸".freeze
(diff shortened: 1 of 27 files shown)


lib/test_prof/recipes/rspec/let_it_be.rb
```diff
@@ -22,7 +22,11 @@ def modules
     # Use uniq prefix for instance variables to avoid collisions
     # We want to use the power of Ruby's unicode support)
     # And we love cats!)
-    PREFIX = "@😸".freeze
+    if RUBY_ENGINE == 'jruby'
+      PREFIX = "@😸 ".freeze
```
sponomarev: LOL. What is the difference?

palkan: I'm trying to figure out; it works only with extra spaces in JRuby irb, but haven't tried running from a file

palkan: Nope, nothing works(

PR #20, opened 2017-09-16
Fix Tips and Tricks anchor

basmoura:
The current anchor for Tips and Tricks links is broken, this PR fix this issue


palkan:
Thanks!

Issue #26, opened 2017-09-26
Order of truncating tables in after(:suite) for AnyFixture


AlexVPopov:
Say you have verious levels of nesting of models:

```ruby
class Foo < ApplicationRecord
  has_many :bars, dependent: :destroy
end

class Bar < ApplicationRecord
  belongs_to :foo
  has_many :bazs, dependent: :destroy
end

class Baz < ApplicationRecord
  belongs_to :bar
end
```

The logical thing to do is to first delete all the `bazs`, then the `bars` and finally the `foos`. If this order is not respected, one might potentially get an error like this one:

```ruby
ActiveRecord::InvalidForeignKey:
  PG::ForeignKeyViolation: ERROR:  update or delete on table "products" violates foreign key constraint "fk_from_subscriptions_to_products" on table "subscriptions"
  DETAIL:  Key (id, facility_id)=(1, 1) is still referenced from table "subscriptions".
  :
              DELETE FROM "products"
```

What do you think? Is it feasible to build some logic for the order of the deletion of tables?

sponomarev:
The tables will be cleaned in the reverse order of entities creation which is logical and has to cover all the cases. It was fixed in #2. Do you see any edge cases and provide a way how to reproduce them?

palkan:
@AlexVPopov 

>  reproduce my edge case into a sample app.

It should be easy to write a reproduction script using our specs (see https://github.com/palkan/test-prof/tree/master/spec/bugs). Feel free to add a failing reproduction spec and propose a PR.

AlexVPopov:
O.K., I'll try and reopen this, thank you!

AlexVPopov:
@palkan @sponomarev O.K., I caught my bug, so here is what was happening:

My baz context looked like this:

```ruby
shared_context 'baz' do
  before(:all) do
    another_bar = Fabricate(:baz, foo: @foo)

    @baz = TestProf::AnyFixture.register(:baz) do
      Fabricate(:baz, foo: @foo, bar: another_bar)
    end
  end

  let(:baz) { foo.bazs.find(@baz.id) }
end
```

and in one spec file I had:

```
include_context 'foo'
include_context 'baz'
```

At this point `tables_cache.keys #=> ['foos', 'bazs']`.

In another spec file I finally had:

```ruby
include_context 'bar'
```

[...]

PR #27, opened 2017-09-26
Update AnyFixture guide with 'Caveats' section

AlexVPopov:
Relates to #26 

The purpose of this request is to document unintended side-effects of creating external records within an `AnyFixture` block.

I updated the `AnyFixture` guide with a *Caveats* section.

Sorry if my description is too verbose or unclear. Please, point in which way the explanation could be improved. Thank you.

### DIFF
--- a/guides/any_fixture.md
+++ b/guides/any_fixture.md
@@ -58,3 +58,78 @@ require "test_prof/recipes/rspec/any_fixture"
 ```
 
 Now you can use `TestProf::AnyFixture` in your tests.
+
+## Caveats
+
+`AnyFixture` cleans tables in the reversed order, in which they were registered. This
+means, that if you register a fixture, which references a not-yet-registed table, a
+foreign-key violation error *might* occur. An example is worth more than 1000
+words:
+
+```ruby
+class Author < ApplicationRecord
+  has_many :articles
+end
+
+class Article < ApplicationRecord
+  belongs_to :author
+end
+```
+
+The usual usage of would be:
+
+1. Register the shared contexts:
+
+```ruby
+RSpec.shared_context "author" do
+  before(:all) do
+    @author = TestProf::AnyFixture.register(:author) do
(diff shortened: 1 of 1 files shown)


guides/any_fixture.md
```diff
@@ -58,3 +58,78 @@ require "test_prof/recipes/rspec/any_fixture"
...
+
+2. Include the contexts into a spec file:
+
+```ruby
+include_context 'author'
+include_context 'article'
+```
+
+If one forgets to include the `author` context or includes it after the `article` context,
+an error will be raised. At the end of the suite, first the `articles` and then the
```
palkan: That's not true, actually. If we don't include the "author" context (or include it after) there is no `@author` during the execution of the "article" context, so it's `nil`.

Then either `create(:article, author: nil)` fails (if there is a validation) or nothing bad happens, 'cause `@article` wouldn't have a `author_id`.

AlexVPopov: @palkan I will check tomorrow, but if I am not mistaken, if I refence `@author`, which doesn't exist yet, I get an error (maybe because in Rails 5 belongs_to is required by default, however, I don't remember what the error exactly was). I'll check tomorrow and I'll make a new PR.

AlexVPopov: What you say sounds quite logical.

palkan: Yep, Rails 5 `belongs_to` adds `required: true` by default.

I've updated the guide and left the second example only.

guides/any_fixture.md
```diff
@@ -58,3 +58,78 @@ require "test_prof/recipes/rspec/any_fixture"
...
+`authors` table will be cleaned.
+
+Here is another example of the `artcile` shared context:
```
sponomarev: Typo

palkan: Thnx. I've already fixed typos in master)

Issue #21, opened 2017-09-21
EventProf Minitest support


palkan:
Add Minitest integration support to EventProf.

### Tracking

I propose to use [minitest-reporters](https://github.com/kern/minitest-reporters) for tests tracking purposes (at least for now). It works pretty similar to RSpec reporter listeners, so could be easily integrated with EventProf core functionality.

### Configuration

We should support both env variables and CLI options (thus, we have to provide a Minitest plugin (see [Rails plugin](https://github.com/rails/rails/blob/master/railties/lib/minitest/rails_plugin.rb), for example)):

```
# that should work
EVENT_PROF=sql.active_record rake test

# and that too
ruby my_test.rb --event_prof=sql.active_record
```

IDolgirev:
Hi!
Well, I believe I'm quite close to finalize this but I've faced with problem to get location of example group for proper report results.

Technically, any Minitest reporter receives such group as a Class object and example name as String in prerecord hook. I can track down location for examples via `source_location`, but can't find a robust solution to get the same for group definition.

Possible solution is to get location of the first example of the group (probably we can subtract one line and believe that there is no blank line between group definition and example definition ) and use it as a group location.
Another one is to explicitly show in group location something like 'FILE: around line XX', where XX is the location of the first group example as well.

Any thoughts or ideas?
Thanks in advance!

palkan:
Hi!

I think, we can just use a file path without the line number for a group location. WDYT?

IDolgirev:
I like such simplicity :)

palkan:
Closed by #29

PR #29, opened 2017-09-29
EventProf Minitest integration

IDolgirev:
It implements Minitest integration for EventProf (#21) 

Custom Minitest plugin and reporter were added
Integration specs (based on the same Rspec ones) were added

If there are no critical comments about I'll update EventProf guide

### DIFF
--- a/lib/test_prof/event_prof/minitest.rb
+++ b/lib/test_prof/event_prof/minitest.rb
@@ -7,21 +7,32 @@ module Minitest
   class EventProfReporter < AbstractReporter # :nodoc:
     include TestProf::Logging
 
-    def initialize(options)
+    attr_accessor :io
+
+    def initialize(io = $stdout, options = {})
+      @io = io
       @profiler = configure_profiler(options)
       @formatter = TestProf::EventProf::MinitestFormatter.new(@profiler)
       @current_group = nil
       @current_example = nil
     end
 
+    def start; end
+
     def prerecord(group, example)
       change_current_group(group, example) unless @current_group
       track_current_example(group, example)
     end
 
+    def before_test(test)
+      prerecord(test.class, test.name)
+    end
+
     def record(*)
       @profiler.example_finished(@current_example)
     end
+
+    def after_test(*); end
 
     def report
       @profiler.group_finished(@current_group)
@@ -39,7 +50,7 @@ module Minitest
       end
 
       @current_example = {
-        name: example.gsub(/^test_\d+_/, ''),
+        name: example.gsub(/^test_(?:\d+_)?/, ''),
         location: File.expand_path(location(group, example).join(':')).gsub(Dir.getwd, '.')
       }
 
@@ -56,7 +67,8 @@ module Minitest
     end
 
     def location(group, example)
-      name = group.public_instance_methods(false).find { |mtd| mtd.to_s == example }
+      suite = group.public_instance_methods.select { |mtd| mtd.to_s.match /^test_/ }
+      name = suite.find { |mtd| mtd.to_s == example }
       group.instance_method(name).source_location
     end
 
--- a/lib/test_prof/event_prof/minitest.rb
+++ b/lib/test_prof/event_prof/minitest.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require 'test_prof/logging'
 require 'test_prof/event_prof/formatters/minitest'
 
 module Minitest
@@ -8,7 +9,7 @@ module Minitest
 
     def initialize(options)
       @profiler = configure_profiler(options)
-      @formatter = MinitestFormatter.new(@profiler)
+      @formatter = TestProf::EventProf::MinitestFormatter.new(@profiler)
       @current_group = nil
       @current_example = nil
     end
--- a/lib/test_prof/event_prof/minitest.rb
+++ b/lib/test_prof/event_prof/minitest.rb
@@ -62,7 +62,10 @@ module Minitest
 
     def configure_profiler(options)
       TestProf::EventProf.configure do |config|
-        config.event = options[:event_prof]
+        config.event = options[:event]
+        config.rank_by = options[:rank_by] if options[:rank_by]
+        config.top_count = options[:top_count] if options[:top_count]
+        config.per_example = options[:per_example] if options[:per_example]
       end
 
 			TestProf::EventProf.build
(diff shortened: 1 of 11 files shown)


lib/test_prof/event_prof/minitest.rb
```diff
@@ -1,3 +1,73 @@
...
+        name: group.name,
+        location: File.expand_path(location(group, example).first).gsub(Dir.getwd, '.')
+      }
+
+      @profiler.group_started(@current_group)
+    end
+
+    def location(group, example)
+      name = group.public_instance_methods(false).find { |mtd| mtd.to_s == example }
+      group.instance_method(name).source_location
```
palkan: I've tried to run it with sample Rails app and got the following:

```
Run options: --event-prof=sql.active_record --seed 5773

# Running:

..../usr/local/lib/ruby/gems/2.4.0/bundler/gems/test-prof-3b646bac1b96/lib/test_prof/event_prof/minitest.rb:60:in `instance_method': nil is not a symbol nor a string (TypeError)
	from /usr/local/lib/ruby/gems/2.4.0/bundler/gems/test-prof-3b646bac1b96/lib/test_prof/event_prof/minitest.rb:60:in `location'
	from /usr/local/lib/ruby/gems/2.4.0/bundler/gems/test-prof-3b646bac1b96/lib/test_prof/event_prof/minitest.rb:52:in `change_current_group'
	from /usr/local/lib/ruby/gems/2.4.0/bundler/gems/test-prof-3b646bac1b96/lib/test_prof/event_prof/minitest.rb:38:in `track_current_example'
	from /usr/local/lib/ruby/gems/2.4.0/bundler/gems/test-prof-3b646bac1b96/lib/test_prof/event_prof/minitest.rb:19:in `prerecord'
```

I think, this is something Rails specific but we should care.

Example app is here https://github.com/palkan/th-dummy/tree/test-prof-minitest

palkan: Also, FYI https://github.com/rails/rails/issues/29469 (maybe, it worth mentioning this issue in the guide).

IDolgirev:
Thanks for the review!
I've already fixed that issue, but found one more little bug I want to fix before update the PR.
Do you have some suggestions about version in changelog?

palkan:
@IDolgirev 

> Do you have some suggestions about version in changelog?

Let's use "master"; I'll update it before releasing.

IDolgirev:
Thanks, got it!

lib/test_prof/logging.rb
```diff
@@ -13,7 +13,7 @@ def log(level, msg)
     end
 
     def build_log_msg(level, msg)
-      colorize(level, "[TEST PROF #{level.to_s.upcase}] #{msg}")
+      colorize(level, "\n[TEST PROF #{level.to_s.upcase}] #{msg}")
```
palkan: What's the reason for the change? I think, it's better to control newlines in-place and do not hide within a private method.

IDolgirev: Yep, you're right - it was just a workaround for default Minitest reporter, where progress line concatenates with EventProf message. I'll remove it.

lib/test_prof/event_prof/minitest.rb
```diff
@@ -37,7 +38,7 @@ def after_test(*); end
     def report
       @profiler.group_finished(@current_group)
       result = @formatter.prepare_results
-
+      puts "\n"
```
palkan: Bad idea. We should output everything using `log`, 'cause (in theory) it can write to file.
Just prepend the `result` with `"\n"` (i.e. `log :info, "\n#{result}"`).

IDolgirev:
Frankly, I don't know what I was thinking about at that moment. Sorry for such a silly mistake :(

Also there is one more thing I need to mention - even all specs are green I think that EventProf guide need to be updated a little bit with some caveat about minitest-reporters. When the user unintentionally run specs without prepending them with `bundle exec`, he (or she) will get an error, because Minitest initialize all its plugins from `minitest/*_plugin.rb` of the every entry in the $LOAD_PATH, and minitest-reporters as just installed gem is available there. And when this gem is not properly configured as I mentioned in the guide, some hooks around the test are become unavailable which causes the error. 
At the moment I suggest to update the guide with that, but will try to investigate it deeper. WDYT?

lib/test_prof/event_prof/minitest.rb
IDolgirev: Actually, it won't help with the concatenation issue, because we need new line before log message itself, but not before the results. But as I wrote below I agree that it was a dumb mistake to didn't keep in mind a possibility to write results to the file :)

palkan:
> At the moment I suggest to update the guide with that, but will try to investigate it deeper

Sounds good!

palkan:
@IDolgirev Thank you for your help! I need your full name in Russian for CultOfMartians and your Twitter nickname (if any) for the announcement.

IDolgirev:
You're welcome! Glad to help :)
My name in Russian is `Илья Долгирев`  and Twitter nickname is `@Dolgirev`

palkan:
@IDolgirev Take a look, please: wrong event count https://travis-ci.org/palkan/test-prof/jobs/282329413

IDolgirev:
Thanks, already trying to reproduce it locally.

As you see the whole group is missing in the results.

I'm not quite good at multithreading and probably need some advices but at the first glance the reason is inside `report` hook when the notification to finish last group is send to profiler it should be handled first, and only then we can use results.
Is it possible to get wrong order of execution of lines inside the same method due to the nature of JRuby?

palkan:
> Is it possible to get wrong order of execution of lines inside the same method due to the nature of JRuby?

Yep, that may be the case. Not sure, but maybe Minitest automatically runs in multithreaded mode in JRuby.

IDolgirev:
Well, I was totally unlucky yesterday so I didn't even get a chance to reproduce the error. Single spec, whole suite, random order, specific seed from Travis' failed job - nothing :(
Maybe you have some tricks how to deal with such type of errors?

palkan:
Have you tried JRuby?)

Let's not spend too much time; I will try to investigate the next this problem occur.

Thank you!

IDolgirev:
In that case it was definitely JRuby because my machine started to work like any Dyson devices :)
And yes, I've checked it via `htop` :)
OK, feel free to ping me if you need some help with that next time.

PR #38, opened 2017-10-07
Factory Doctor Minitest integration

IDolgirev:
It implements Minitest integration for Factory Doctor.

Minitest related fixture and integration spec were added.
UPD: Feature to skip specific examples was added 
Factory Doctor Minitest Reporter was added.
EventProf Minitest plugin was slightly refactored and turned out to TestProf Minitest Plugin.
EventProf Reporter was refactored.

There is a minor issue when we get double log message about enabling Factory Doctor via env variable
I tracked it down to [https://github.com/palkan/test-prof/blob/master/lib/test_prof/factory_doctor.rb#L139-L141](https://github.com/palkan/test-prof/blob/master/lib/test_prof/factory_doctor.rb#L139-L141), and I think it could be easily moved to RSpecListener.(UPD: Fixed with suggested scenario)


palkan:
Wow! Great; I'll be able to take a look at it in the end of the next week (ping me if I won't do that).

It would be great if you reduce the number of commits (i.e. rebase), btw)

IDolgirev:
OK, sure!

So, if we have some time I'd like to ask your thoughts about possible solution for ignoring examples in Minitest.
The first option that came to my head is to use naming convention for example, something like 'test_name_of the_example_fd_ignore'. I believe it will be simple to implement, but I'm not totally sure about the idea of polluting names of the examples.

The second one is to implement custom `fd_ignore' which will be used like a native `skip`. It seems more suitable, IMO.

WDYT?

palkan:
> The second one is to implement custom fd_ignore which will be used like a native skip. It seems more suitable, IMO.

I like this idea

IDolgirev:
Hi!
It's a tiny reminder about review you've asked for.
And it's not me who've failed the build - it was Travis! :)

IDolgirev:
Hi!
I'm sorry for disturbing you, but do I have to do something else before the start of the review? :)

lib/test_prof/event_prof/minitest.rb
```diff
@@ -1,90 +1,73 @@
...
+        @profiler.example_finished(@current_example)
+      end
 
-    def track_current_example(group, example)
-      unless @current_group[:name] == group.name
+      def report
         @profiler.group_finished(@current_group)
-        change_current_group(group, example)
+        result = @formatter.prepare_results
+        puts "\n"
```
palkan: ??

lib/test_prof/factory_doctor/minitest.rb
```diff
@@ -1,3 +1,91 @@
...
+  module TestProf
+    class FactoryDoctorReporter < BaseReporter # :nodoc:
+      using ::TestProf::FloatDuration
+      using ::TestProf::StringStripHeredoc
+
+      SUCCESS_MESSAGE = 'FactoryDoctor says: "Looks good to me!"'.freeze
+
+      def initialize(io = $stdout, options = {})
+        super
+        @doctor = configure_doctor
```
palkan: What's the reason of storing constant in an instance variable?

spec/support/integration_helpers.rb
```diff
@@ -14,7 +14,7 @@ def run_rspec(path, chdir: nil, success: true, env: {})
   def run_minitest(path, chdir: nil, success: true, env: {})
     output, status = Open3.capture2(
       env,
-      "ruby #{path}_fixture.rb",
+      "ruby #{path}_fixture.rb #{env['TESTOPTS']}",
```
palkan: Shouldn't it work without passing explicitly as arg? I mean, having `TESTOPTS` env variable should just work

IDolgirev: It will work perfectly when running test via `rake`. Running test via Ruby interpreter accepts test name as CLI option. That's why it's here.

palkan: Ok

palkan:
Please, rebase

IDolgirev:
Sure, will do little bit later because I'm on dad's duty for now :)
Do I need to squash all into one commit?

palkan:
> Do I need to squash all into one commit?

I'd prefer to have several commits, just fix up a little bit.

IDolgirev:
Done!

palkan:
Cool! Thank you!

IDolgirev:
Glad to help! :)

PR #42, opened 2017-10-11
FactoryProf: Do not display error message when Fabrication gem is not installed

### LINKED ISSUE #40: [TEST PROF ERROR] appears when running FactoryProf, but results are correct
**Ruby 2.3.4**

**Rails 4.2.0, RSpec 3.1.0**

**TestProf 0.4.0**

After `FPROF=1 rspec` run I've got the following output:

```
[TEST PROF INFO] FactoryProf enabled (simple mode)
[TEST PROF ERROR]
```

However, when tests are finished, I have proper output about factories usage.

It also appears with `FPROF=flamegraph rspec`

Shkrt:
Fixes #40. Remove from logs confusing empty error message caused by missing fabrication gem.

### DIFF
--- a/lib/test_prof/factory_prof/factory_builders/fabrication.rb
+++ b/lib/test_prof/factory_prof/factory_builders/fabrication.rb
@@ -10,9 +10,9 @@ module TestProf
       class Fabrication
         # Monkey-patch Fabrication
         def self.patch
-          TestProf.require 'fabrication', ""
-          ::Fabricate.singleton_class.prepend(FabricationPatch) if
-            defined?(::Fabrication)
+          TestProf.require 'fabrication' do
+            ::Fabricate.singleton_class.prepend(FabricationPatch)
+          end
         end
 
         def self.track(factory, &block)
--- a/lib/test_prof/factory_prof/factory_builders/factory_girl.rb
+++ b/lib/test_prof/factory_prof/factory_builders/factory_girl.rb
@@ -10,8 +10,9 @@ module TestProf
       class FactoryGirl
         # Monkey-patch FactoryGirl
         def self.patch
-          ::FactoryGirl::FactoryRunner.prepend(FactoryGirlPatch) if
-            defined?(::FactoryGirl)
+          TestProf.require 'factory_girl' do
+            ::FactoryGirl::FactoryRunner.prepend(FactoryGirlPatch)
+          end
         end
 
         def self.track(strategy, factory, &block)
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@
 /tmp/
 *.gem
 gemfiles/*.lock
+spec/integrations/fixtures/rspec/tmp/
+spec/integrations/fixtures/rspec/gemfiles/*.lock
--- a/lib/test_prof.rb
+++ b/lib/test_prof.rb
@@ -47,11 +47,11 @@ module TestProf
 
     # Require gem and shows a custom
     # message if it fails to load
-    def require(gem_name, msg)
+    def require(gem_name, msg = nil)
       Kernel.require gem_name
       block_given? ? yield : true
     rescue LoadError
-      log :error, msg
+      log(:error, msg) if msg
       false
     end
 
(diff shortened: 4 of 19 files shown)


palkan:
I don't think that fix #40. What we want is to not show any message when one of the libraries is not loaded, i.e. when you're running FPROF but do not create factories at all.

My suggestion is to:

- enhance `TestProf.require(gem, msg)` to handle `nil` as "do not log anything" and use it as default

- use `TestProf.require("x")` for both FactoryGirl and Fabrication.

Thus when no library is found we just print at the end that there is no factories data.

And we need an integration test for that.

Shkrt:
It seems that there is no way to do integration testing to reflect the case when some of the gems are not installed. Of course, we can mock `Kernel.require` method, but it gets executed before our test suite, so this does not make sense. Maybe, we should unit test `TestProf#require` method instead?

palkan:
We can try to tweak with bundler options for integrations tests (e.g. set specific `BUNDLE_GEMFILE` or smth  like `Bundler#with_clean_env`).

Unit-testing `TestProf#require` is a good idea too.

Shkrt:
Integration and unit spec were added. I had to introduce separate gemfiles for fixtures, which reside in `spec/integrations/fixtures/rspec/gemfiles`. Also, I had to commit lockfiles for "railsmaster" gemfiles, because Bundler fails to check out gems from github, if being run from inside a ruby script, for unknown reasons. If I don't commit them, "ruby-head" build always fails (BTW it is in `allow_failures` section in travis config)

palkan:
Good job!

Unfortunately, locking rails version makes this configuration useless, 'cause we want to follow the latest Rails master and not the specific version.

What is the Bundler error when there is no lock file?

Shkrt:
> What is the Bundler error when there is no lock file?
 
`The git source git://github.com/rails/arel.git is not yet checked out. Please run 'bundle install' before trying to start your application (Bundler::GitError)`

palkan:
Well, let's try this https://github.com/rdp/specific_install and install gems manually for `railsmaster` gemfile. Yes, it's becoming too complex(

I think, that would require us to add a special script to install gems and run it only for `railsmaster` (by adding a special env var to the matrix):

```
before_install
  - gem install bundler -v 1.13.6
  - if [ "${RAILS_MASTER}" = "true" ]; "./gemfiles/prepare_rails_master.sh" ; fi 

...
matrix:
  include:
    - rvm: ruby-head
      gemfile: gemfiles/railsmaster.gemfile
      env:
        RAILS_MASTER=true
```

Shkrt:
It seems that everything is working now. Unfortunately, rdp/specific_install wasn't of any help, because it has its own [limitations](https://github.com/rdp/specific_install/issues/43) and when I even ended up patching it during travis build, it did not solve the problem.
Finally, we had to combine `Bundler.with_clean_env` and `ENV['BUNDLE_GEMFILE']` (which is not, of course, cleared when using `with_clean_env`).

lib/test_prof/factory_prof/factory_builders/fabrication.rb
```diff
@@ -10,7 +10,7 @@ module FactoryBuilders
       class Fabrication
         # Monkey-patch Fabrication
         def self.patch
-          TestProf.require 'fabrication', ""
+          TestProf.require 'fabrication'
           ::Fabricate.singleton_class.prepend(FabricationPatch) if
```
palkan: We can move this line into `TestProf.require` block and avoid checking for the constant.

lib/test_prof/factory_prof/factory_builders/factory_girl.rb
```diff
@@ -10,6 +10,7 @@ module FactoryBuilders
       class FactoryGirl
         # Monkey-patch FactoryGirl
         def self.patch
+          TestProf.require 'factory_girl'
           ::FactoryGirl::FactoryRunner.prepend(FactoryGirlPatch) if
```
palkan: The same here

palkan:
Just a few small changes and we're done)

Shkrt:
Done.

palkan:
Great! Thanks a lot!

palkan:
FYI https://github.com/palkan/test-prof/commit/49452ce9424260e0a905a318973e402082df8bef

Looks like an easier way) And works perfectly with different CI envs.

PR #46, opened 2017-10-24
Migration from FactoryGirl to FactoryBot

### LINKED ISSUE #45: 🤖FactoryBot killed 🙅FactoryGirl
Since FactoryGirl has been renamed to [FactoryBot](https://github.com/thoughtbot/factory_bot) we have to support both names now (for users still using good old FactoryGirl).

I propose to handle this duality by using our own `TestProf::FactoryBot` (just an alias for `FactoryGirl` or `FactoryBot`) everywhere in codebase.

We should also use `factory_bot` by default in our tests and add a separate Travis gemfile for `factory_girl`.

Shkrt:
Fixes #45 

Wrap-up:
* Each travis build now has two gemfiles - one with `factory_bot` and another one with `factory_girl`
* Introduced `TestProf::FactoryBot` constant, which replaces `FactoryGirl` and points either to `FactoryBot` or `FactoryGirl`, if one of them is installed
* Error message in `test-prof/lib/test_prof/event_prof/custom_events/factory_create.rb` file replaced with more generic one

### DIFF
--- a/lib/test_prof/event_prof/custom_events/factory_create.rb
+++ b/lib/test_prof/event_prof/custom_events/factory_create.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "test_prof/ext/string_strip_heredoc"
+require "test_prof/factory_bot"
 
 using TestProf::StringStripHeredoc
 
@@ -18,7 +19,7 @@ module TestProf::EventProf::CustomEvents
     class << self
       def setup!
         @depth = 0
-        FactoryGirl::FactoryRunner.prepend RunnerPatch
+        TestProf::FactoryBot::FactoryRunner.prepend RunnerPatch
       end
 
       def track(factory)
@@ -45,11 +46,11 @@ end
 
 TestProf.activate('EVENT_PROF', 'factory.create') do
   if TestProf.require(
-    'factory_girl',
+    'factory_bot',
     <<-MSG.strip_heredoc
-      Failed to load FactoryGirl.
+      Failed to load FactoryBot.
 
-      Make sure that "factory_girl" gem is in your Gemfile.
+      Make sure that "factory_bot" gem is in your Gemfile.
     MSG
   )
     TestProf::EventProf::CustomEvents::FactoryCreate.setup!
--- a/lib/test_prof/event_prof/custom_events/factory_create.rb
+++ b/lib/test_prof/event_prof/custom_events/factory_create.rb
@@ -45,14 +45,15 @@ module TestProf::EventProf::CustomEvents
 end
 
 TestProf.activate('EVENT_PROF', 'factory.create') do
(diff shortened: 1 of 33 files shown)


lib/test_prof/event_prof/custom_events/factory_create.rb
```diff
@@ -1,6 +1,10 @@
 # frozen_string_literal: true
 
 require "test_prof/ext/string_strip_heredoc"
+require "test_prof/factory_bot"
+require "test_prof/logging"
+
+extend TestProf::Logging
```
palkan: Let's not pollute global namespace) Can't we just use `TestProf.log(...)`?

lib/test_prof/event_prof/custom_events/factory_create.rb
```diff
@@ -44,14 +48,20 @@ def track(factory)
...
-  if TestProf.require(
-    'factory_girl',
-    <<-MSG.strip_heredoc
-      Failed to load FactoryGirl.
-
-      Make sure that "factory_girl" gem is in your Gemfile.
-    MSG
-  )
-    TestProf::EventProf::CustomEvents::FactoryCreate.setup!
+  loaded = %w[factory_bot factory_girl].inject(false) do |_acc, elem|
```
palkan: Let's make it a little bit shorter and more readable (IMO):

```ruby
if %w[factory_bot factory_girl].find { |name| TestProf.require(name) }
  TestProf::EventProf::CustomEvents::FactoryCreate.setup!
else
  TestProf.log(...)
end
```

lib/test_prof/factory_bot.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module TestProf
+  CONSTANT_NAMES = { 'factory_bot' => '::FactoryBot', 'factory_girl' => '::FactoryGirl' }.freeze
+
+  CONSTANT_NAMES.each_key do |name|
```
palkan: We can use `find` trick here too:

```ruby
CONSTANT_NAMES.find do |name, cname|
  TestProf.require(name) do
     TestProf::FactoryBot = Object.const_get(cname)
  end
end
```

lib/test_prof/factory_bot.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module TestProf
+  CONSTANT_NAMES = { 'factory_bot' => '::FactoryBot', 'factory_girl' => '::FactoryGirl' }.freeze
```
palkan: What about changing nam to `FACTORY_GIRL_NAMES` and using it in `FactoryCreate` too?

Shkrt:
Done.

palkan:
Great! Thank you!

Let's add a changelog entry and we're done.

PR #51, opened 2017-11-10
Сorrect travis build matrix syntax

Shkrt:
Recently I  have discovered, that 2 of 9 jobs are not actually run by Travis due to a syntax error in `.travis.yml`. However, somehow, those jobs are always marked as "successful" and errors seen only after researching the job log.  This syntax was initially found in [Travis docs](https://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix), but apparently, it doesn't get parsed correctly in some scenarios.


palkan:
Thanks!

PR #52, opened 2017-11-30
Replace logo styles with attributes

askd:


palkan:
Let's use this updated logo in Readme.md

PR #57, opened 2017-12-24
fix RubyProf Printers

rabotyaga:
fixes support of RubyProf printers:
 - FlatPrinterWithLineNumbers - minor typo fixed
 - CallTreePrinter - needs different arguments
+ added support for MultiPrinter, needs same arguments as CallTreePrinter

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## master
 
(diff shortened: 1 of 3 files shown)


palkan:
Thanks!

Let's add a Changelog entry.

Issue #80, opened 2018-06-20
Issue with create_default and traits


dgilperez:
I have this factory, with a trait that allows changing the association general definition like this: 

```ruby
FactoryBot.define do
  factory :idea do
    association :status, :default

    trait :bla do
      association :status, :bla
    end
  end
end 
```

When I use FactoryDefault to define a default `status` like this in a shared context:

```ruby
shared_context 'My context' do  
  let_it_be(:status) { create :status }

  before do
    set_factory_default :status, status
  end
end
```

And write my spec: 

```ruby
include_context 'My context'

it do
  create :idea #=> is created with the default :status from shared context
  create :idea, :bla #=> is ALSO created with the default :status from shared context
end
```

So in practice `trait :bla` is not used at all, `set_factory_default` is overriding that. This was, at least, surprising and unexpected. I expected the trait behaviour to preveal, since it is explicitly declaring the association.

Is this intented behaviour or is this a 🐛? 

**Ruby Version:** 2.3.7
**Framework Version:** Rspec 3.6
**TestProf Version:** 0.5.0

palkan:
> Is this intented behaviour or is this a 🐛?

It's an intended behaviour. `FactoryDefault` doesn't take into account traits.

I think, we can add this functionality in the following way:

```ruby
# make this record default for the specified traits
create_default(:idea, :my_trait, **params, preserve_traits: true)
# or
set_factory_default(:idea, :my_trait, obj)

# we can set `preserve_traits` globally
TestProf::FactoryDefault.preserve_traits = true

# by default should be false for backward compatibility
```

palkan:
@dgilperez Would you like to propose a PR, btw?

dgilperez:
So I was thinking about the proposed interface, and I was thinking more in the other direction: 

My thinking is that having as we have several different traits per factory, playing with associations, your proposal of defining different defaults for different traits will end up in the need to define one default per trait, and the need to keep remembering that (I will keep forgetting for sure 😄). I'd prefer a way to get my expected behaviour above. 

Something like _use default factory, unless you define it explicitly in a trait_. 

```
# default false for backward compatibility
TestProf::FactoryDefault.preserve_traits = true 

set_factory_default(:idea, obj)

create :idea_thing # =>  sets obj as default
create :idea_thing, :trait_that_changes_idea_association # =>  does not set obj as default
```

I'd definitely like to contribute. But I will not be able until, at least, August. If someone wants to get ahead, that'd be great as well.

[...]

PR #82, opened 2018-06-24
Add possibility to preserve traits

Mordorreal:
Now the user can set FactoryDefault.preserve_traits to true and the
factory default will fallback to the default factory behaviour.

Ref: #80

I faced some problem with local test running. Got for every spec:
```
NameError:
  uninitialized constant TestProf::FactoryBot
# ./spec/support/ar_models.rb:42:in `<top (required)>'
# ./spec/spec_helper.rb:12:in `block in <top (required)>'
# ./spec/spec_helper.rb:12:in `each'
# ./spec/spec_helper.rb:12:in `<top (required)>'
# ./spec/test_prof_spec.rb:3:in `<top (required)>'
```
 I'll be very appreciated for your help with this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## master
 
+- Add `FactoryDefault.preserve_traits` to preserve traits. ([@mordorreal][])
+
 - Add `EventProf.monitor` to instrument arbitrary methods. ([@palkan][])
 
 Add custom instrumetation easily:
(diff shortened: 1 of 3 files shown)


palkan:
Implemented in #83.

palkan:
> I faced some problem with local test running. Got for every spec:

Does your `Gemfile.lock` contain `factory_bot`?

PR #83, opened 2018-06-26
[Fix #80] Ability to preserve traits

### LINKED ISSUE #80: Issue with create_default and traits

Vasfed:
Fixes #80.

This adds option to `TestProf::FactoryDefault` to revert to default behaviour when factories have traits for association.

### DIFF
--- a/lib/test_prof/factory_default.rb
+++ b/lib/test_prof/factory_default.rb
@@ -10,14 +10,14 @@ module TestProf
     module DefaultSyntax # :nodoc:
       def create_default(name, *args, &block)
         options = args.extract_options!
-        preserve = options.delete(:preserve_traits) || FactoryDefault.preserve_traits
+        preserve = options.delete(:preserve_traits)
 
         obj = TestProf::FactoryBot.create(name, *args, options, &block)
-        set_factory_default(name, obj, preserve)
+        set_factory_default(name, obj, preserve_traits: preserve)
       end
 
-      def set_factory_default(name, obj, preserve_traits = nil)
-        FactoryDefault.register(name, obj, preserve_traits)
+      def set_factory_default(name, obj, preserve_traits: nil)
+        FactoryDefault.register(name, obj, preserve_traits: preserve_traits)
       end
     end
 
@@ -32,27 +32,28 @@ module TestProf
         TestProf::FactoryBot::Strategy::Stub.prepend StrategyExt
 
         @store = {}
-        @store_preserve_traits = {}
         # default is false to retain backward compatibility
         @preserve_traits = false
       end
 
-      def register(name, obj, preserve_traits = nil)
-        unless FactoryDefault.preserve_traits
-          @store_preserve_traits[name] ||= true if preserve_traits
-        end
-        store[name] = obj
+      def register(name, obj, **options)
+        options[:preserve_traits] = true if FactoryDefault.preserve_traits
+        store[name] = { object: obj, **options }
+        obj
       end
 
-      def get(name, _traits = nil)
-        store[name]
+      def get(name, traits = nil)
+        record = store[name]
+        return unless record
+
+        if traits && !traits.empty?
+          return nil if FactoryDefault.preserve_traits || record[:preserve_traits]
+        end
+        record[:object]
       end
 
       def exists?(name, traits = nil)
-        if traits && !traits.empty?
-          return false if FactoryDefault.preserve_traits || @store_preserve_traits[name]
-        end
-        store.key?(name)
+        get(name, traits) && true || false
       end
 
       def remove(name)
@@ -61,7 +62,6 @@ module TestProf
 
       def reset
         @store.clear
-        @store_preserve_traits.clear
       end
 
       private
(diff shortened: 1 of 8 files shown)


lib/test_prof/factory_default.rb
```diff
@@ -9,18 +9,21 @@ module TestProf
...
+        options = args.extract_options!
+        preserve = options.delete(:preserve_traits) || FactoryDefault.preserve_traits
+
+        obj = TestProf::FactoryBot.create(name, *args, options, &block)
+        set_factory_default(name, obj, preserve)
       end
 
-      def set_factory_default(name, obj)
-        FactoryDefault.register(name, obj)
+      def set_factory_default(name, obj, preserve_traits = nil)
```
palkan: Let's make it a keyword argument and pass it as `.register(name, obj, options)`. We'll probably have more options in the future.

lib/test_prof/factory_default.rb
```diff
@@ -29,17 +32,26 @@ def init
...
+        # default is false to retain backward compatibility
+        @preserve_traits = false
       end
 
-      def register(name, obj)
+      def register(name, obj, preserve_traits = nil)
+        unless FactoryDefault.preserve_traits
+          @store_preserve_traits[name] ||= true if preserve_traits
+        end
         store[name] = obj
```
palkan: What about storing the options along with the object itself, i.e.:

```ruby
store[name] = { record: obj, **options}
```

spec/integrations/factory_default_spec.rb
```diff
@@ -6,6 +6,6 @@
   specify "RSpec integration", :aggregate_failures do
     output = run_rspec('factory_default')
 
-    expect(output).to include("5 examples, 0 failures")
+    expect(output).to include(" examples, 0 failures")
```
palkan: ?

Vasfed: here we are interested more in the spec passing, thought it is redundant to update test count each time the fixture spec has a test added, restored to explicit count

lib/test_prof/factory_default.rb
Vasfed: First thought was to save some hash allocations, but agree, no point in chasing those several dozens additional hashes, also if we decide to make some default-for-trait-objects they can be stored in the same hash under special keys.

Vasfed:
@palkan please see new commit

lib/test_prof/factory_default.rb
```diff
@@ -32,27 +32,28 @@ def init
...
-        store[name]
-      end
+      def get(name, traits = nil)
+        record = store[name]
+        return unless record
 
-      def exists?(name, traits = nil)
         if traits && !traits.empty?
-          return false if FactoryDefault.preserve_traits || @store_preserve_traits[name]
+          return nil if FactoryDefault.preserve_traits || record[:preserve_traits]
```
palkan: Unnecessary `nil`

lib/test_prof/factory_default.rb
```diff
@@ -32,27 +32,28 @@ def init
...
         end
-        store.key?(name)
+        record[:object]
+      end
+
+      def exists?(name, traits = nil)
```
palkan: Do we use `exists?` anywhere?

Vasfed: No, we don't, can be safely removed - left it for compatibility. It was public and appeared in generated documentation, someone could have used it for some test suite debugging. (sure I doubt it too)

lib/test_prof/factory_default.rb
Vasfed: Enabled rubocop to check for that

lib/test_prof/factory_default.rb
palkan: Let's remove then

Vasfed: Ok, done

palkan:
Thanks! Great job!

palkan:
Hey @Vasfed!

Could you please send me your actual email address to `palkan@evl.ms`? We plan to send some goodies to the Cult participants)

Vasfed:
@palkan Hey! Sent email, e064395fd8a88d6d6c887132d7e1e39035e841a12fc6c5f28cc348ffb4c62b16

PR #104, opened 2018-10-20
Add percentage and total test time to EventPro

### LINKED ISSUE #100: Add percentage and total test time to EventProf
Add a percentage of time spent in the event (from the total test suite/group/example time)

Before:

```
[TEST PROF INFO] EventProf results for factory.create

Total time: 03:07.353
Total events: 7459

Something (./my_spec.rb:21) – 09:16.100 (7 / 3)
...
```

After:

```
[TEST PROF INFO] EventProf results for factory.create

Total time: 03:07.353 of 07:25:232 (42.02%)
Total events: 7459

Something (./my_spec.rb:21) – 09:16.100 (7 / 3) of 12:29:233 (74.13%)
...
```

dmagro:
Added a counter called total_test_time to the `Profiler` class that gets incremented with each example time. This attribute helps the `Profiler`class's caller to know the total test time of all runs. Some other changes are related with the percentage took by the examples. Those changes use the `absolute_run_time` to calculate that percentage and the `EventProfFormatter` and `RSpecListener` methods that involved the reporting were updated. The `run_time` field was added to the group and examples results in order to calculate the same absolute run time for each one of them.

Closes #100

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,17 @@
 
 ## master
 
+- Add a header with the general information on factories usage [#99](https://github.com/palkan/test-prof/issues/99) ([@szemek][])
+
+- Improve test sampling.([@mkldon][])
+
+ ```bash
+ SAMPLE=10 rake test # runs 10 random test examples
+ SAMPLE_GROUPS=10 rake test # runs 10 random example groups
+ ```
+
+ - Extend Event Prof formatter results that includes the absolute run time and percentage of event time percentage [#100](https://github.com/palkan/test-prof/issues/100) ([@dmagro][])
+
 ## 0.7.2 (2018-10-08)
 
 - Add `RSpec/AggregateFailures` support for non-regular 'its' examples. ([@broels][])
@@ -368,3 +379,4 @@ Ensure output dir exists in `#artifact_path` method.
 [@desoleary]: https://github.com/desoleary
 [@rabotyaga]: https://github.com/rabotyaga
 [@Vasfed]: https://github.com/Vasfed
+[@dmagro]: https://github.com/dmagro
--- a/lib/minitest/event_prof_formatter.rb
+++ b/lib/minitest/event_prof_formatter.rb
@@ -26,11 +26,13 @@ module Minitest
       private
 
       def total_results(profiler)
+        time_percentage = time_percentage(profiler.total_time, profiler.absolute_run_time)
+
         @results <<
           <<~MSG
             EventProf results for #{profiler.event}
 
-            Total time: #{profiler.total_time.duration}
+            Total time: #{profiler.total_time.duration} of #{profiler.absolute_run_time.duration} (#{time_percentage}%)
             Total events: #{profiler.total_count}
 
             Top #{profiler.top_count} slowest suites (by #{profiler.rank_by}):
@@ -39,30 +41,46 @@ module Minitest
       end
 
       def by_groups(profiler)
-        profiler.results[:groups].each do |group|
+        result = profiler.results
+        groups = result[:groups]
+
+        groups.each do |group|
           description = group[:id][:name]
           location = group[:id][:location]
+          time = group[:time]
+          run_time = group[:run_time]
+          time_percentage = time_percentage(time, run_time)
 
           @results <<
             <<~GROUP
-              #{description.truncate} (#{location}) – #{group[:time].duration} (#{group[:count]} / #{group[:examples]})
+              #{description.truncate} (#{location}) – #{time.duration} (#{group[:count]} / #{group[:examples]}) of #{run_time.duration} (#{time_percentage}%)
             GROUP
         end
       end
 
       def by_examples(profiler)
-        return unless profiler.results[:examples]
+        result = profiler.results
+        examples = result[:examples]
+
+        return unless examples
         @results << "\nTop #{profiler.top_count} slowest tests (by #{profiler.rank_by}):\n\n"
 
-        profiler.results[:examples].each do |example|
+        examples.each do |example|
           description = example[:id][:name]
           location = example[:id][:location]
+          time = example[:time]
+          run_time = example[:run_time]
+          time_percentage = time_percentage(time, run_time)
 
           @results <<
             <<~GROUP
-              #{description.truncate} (#{location}) – #{example[:time].duration} (#{example[:count]})
+              #{description.truncate} (#{location}) – #{time.duration} (#{example[:count]}) of #{run_time.duration} (#{time_percentage}%)
             GROUP
         end
+      end
+
+      def time_percentage(time, total_time)
+        (time / total_time * 100).round(2)
       end
     end
   end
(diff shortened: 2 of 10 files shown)


lib/minitest/event_prof_formatter.rb
```diff
@@ -42,14 +44,20 @@ def by_groups(profiler)
         profiler.results[:groups].each do |group|
           description = group[:id][:name]
           location = group[:id][:location]
+          time = group[:time]
+          time_percentage = time_percentage(time, profiler.total_test_time)
```
palkan: Here we want to show the percentage of event time to the total **group time** (not the whole suite time).

So, we need to track each example time too.

lib/test_prof/event_prof/profiler.rb
```diff
@@ -40,6 +41,7 @@ def track(time)
 
         @example_time += time
         @example_count += 1
+        @total_test_time += example_time
```
palkan: The `time` passed to `track` method is the event time.

We can try to extend `track` method to accept two arguments: `event_time` (which we track right now) and `total_time` (the total time example ran).

dmagro: I'm not sure how we can capture the total_time. Can you elaborate a little but more?

lib/minitest/event_prof_formatter.rb
dmagro: Quick question, sometimes the example total time can be equal to the total_test_time?

lib/test_prof/event_prof/profiler.rb
palkan: > how we can capture the total_time.

Smth like this:

```ruby
def initialise
  # ...
  @total_run_time = 0.0
end

def group_finished(id)
  group_run_time = TestProf.now - @group_ts

  @total_run_time += group_run_time
  data = { id: id, time: @time, run_time: group_run_time, count: @count, examples: @total_examples }
  @groups << data unless data[rank_by].zero?

  @current_group = nil
end

def example_finished(id)
  @total_examples += 1
  return unless per_example?
        
  data = { id: id, time: @example_time, run_time: TestProf.now - @example_ts, count: @example_count }
  @examples << data unless data[rank_by].zero?
  @current_example = nil
end

def reset_group!
  @time = 0.0
  @count = 0
  @total_examples = 0
  @group_ts = TestProf.now
end

def reset_example!
  @example_count = 0
  @example_time = 0.0
  @example_ts = TestProf.now
end
```

dmagro: I understande the general logic but wouldn't `@time` be the same as the `group_run_time`?

palkan: > wouldn't @time be the same as the group_run_time

Unlikely.

`@time` is the time spending _within an event_, e.g. time spent for database queries.

`run_time` is the whole example/group time, so it is always greater than `@time`, 'cause it contains it

dmagro: Got it

lib/test_prof/event_prof/profiler.rb
```diff
@@ -73,13 +73,18 @@ def example_finished(id)
       end
 
       def results
-        {
+        results = {
           groups: @groups.to_a
         }.tap do |data|
+          data[:groups_total_time] = data[:groups].reduce(0) do |sum, group|
```
dmagro: @palkan could you give some feedback on this approach? I think it's right, but I would like to be sure.

lib/test_prof/event_prof/profiler.rb
dmagro: > > how we can capture the total_time.
> 
> Smth like this:
> 
> ```ruby
> def initialise
>   # ...
>   @total_run_time = 0.0
> end
> 
> def group_finished(id)
>   group_run_time = TestProf.now - @group_ts
> 
>   @total_run_time += group_run_time
>   data = { id: id, time: @time, run_time: group_run_time, count: @count, examples: @total_examples }
>   @groups << data unless data[rank_by].zero?
> 
>   @current_group = nil
> end
> 
> def example_finished(id)
>   @total_examples += 1
>   return unless per_example?
>         
>   data = { id: id, time: @example_time, run_time: TestProf.now - @example_ts, count: @example_count }
>   @examples << data unless data[rank_by].zero?
>   @current_example = nil
> end
> 
> def reset_group!
>   @time = 0.0
>   @count = 0
>   @total_examples = 0
>   @group_ts = TestProf.now
> end
> 
> def reset_example!
>   @example_count = 0
>   @example_time = 0.0
>   @example_ts = TestProf.now
> end
> ```

I think I'm getting this wrong. On the tests I'm always getting small times because there's no real events. Who can I test this on a real situations

palkan: > On the tests I'm always getting small times because there's no real events

Yeah, tests use non-real timings. For real events that shouldn't be the case though.

One way is to add `sleep` statements to https://github.com/palkan/test-prof/blob/master/spec/integrations/fixtures/rspec/event_prof_fixture.rb#L12 (and other fixtures) to make events timings more realistic (and likely change all the value to be less than a second (i.e. 240 -> 0.24)).

palkan: > The change that you spoke before about the track method implies that the event already brings the total_time together with the time

With the above approach we don't need to modify the `track` method

lib/test_prof/event_prof/profiler.rb
```diff
@@ -71,17 +73,27 @@ def example_finished(id)
...
-        {
-          groups: @groups.to_a
+        results = {
+          groups: @groups.to_a,
+          groups_total_time: total_time_for(@groups.to_a)
         }.tap do |data|
           next unless per_example?
 
           data[:examples] = @examples.to_a
+          data[:examples_total_time] = total_time_for(@examples.to_a)
```
palkan: This is not the "total time" we need; we want to compare the event time with the _absolute_ time (maybe, "total" sounds confusing)

lib/test_prof/event_prof/profiler.rb
dmagro: I'll get to it when I have the time, which doesn't come often. Sorry for the delay!

lib/test_prof/event_prof/profiler.rb
palkan: Yep

dmagro:
@palkan could you give it another look?

dmagro:
@palkan can you check if something is missing?

CHANGELOG.md
```diff
@@ -378,3 +378,5 @@ Ensure output dir exists in `#artifact_path` method.
 [@Vasfed]: https://github.com/Vasfed
 [@szemek]: https://github.com/szemek
 [@mkldon]: https://github.com/mkldon
+[@dmagro]: https://github.com/dmagro
```
palkan: Let's add also a change log entry under `master` section in this document

dmagro: @palkan can you check it?

palkan: 👍

palkan:
Thanks for your help!

Issue #133, opened 2019-04-29
before_all hooks


danielwaterworth:
I'm working on speeding up the [discourse](https://github.com/discourse/discourse) tests using `let_it_be`. We need a way to run code that sets up global variables before each before_all block is invoked. I can do this by providing a custom `BeforeAll` adapter within discourse, but if you would like, I could make it more general and submit a PR here. What would you prefer?

palkan:
Hey @danielwaterworth!

> We need a way to run code that sets up global variables before each before_all block is invoked

Yeah, I see (https://github.com/discourse/discourse/pull/7414/files#diff-e11bdaf9116be3911e9557b69e242370R213).

>  I could make it more general and submit a PR here

Yeah, that would be great.
We already have an example where it could be useful: https://github.com/palkan/test-prof/blob/a783083988e6b461f7ad32aa891ad57962a1b97d/lib/test_prof/before_all/adapters/active_record.rb#L15

It could have the following API:

```ruby
Before.configure do |config|
  config.setup_before_all { }
  config.teardown_before_all { }
end
```

palkan:
Closed by #140

PR #135, opened 2019-04-30
Added a hook to trigger an action before each before_all block

danielwaterworth:
#133

API:

    TestProf::BeforeAll.configure |config|
      config.setup_before_all do
        # Code here is run before each before_all block
      end
    end

### DIFF
--- a/lib/test_prof/before_all.rb
+++ b/lib/test_prof/before_all.rb
@@ -18,6 +18,7 @@ module TestProf
       def begin_transaction
         raise AdapterMissing if adapter.nil?
         adapter.begin_transaction
+        config.run_setup
         yield
       end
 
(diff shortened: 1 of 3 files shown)


lib/test_prof/before_all.rb
```diff
@@ -15,9 +15,10 @@ def initialize
     class << self
       attr_accessor :adapter
 
-      def begin_transaction
+      def begin_transaction(ctx)
```
danielwaterworth: @palkan, This bit deserves your attention. I made the hook run in the same context as `before_all` gets in rspec. I think I did the same in minitest, but haven't tested that.

palkan: What is a purpose of passing a context and executing a hook within it?
Isn't `setup_before_all` meant for _global state_ management?

danielwaterworth: @palkan, is there a good reason not to allow access to instance variables?

danielwaterworth: Providing a context means that there's a path forward towards reducing global state.

palkan: > Providing a context means that there's a path forward towards reducing global state.

I see the opposite: by allowing to manipulate the example group state from the global hook could introduce a global dependency adding too many responsibility to `before_all`. That means, that with the ability to _hack_ the context, we can lose the interoperability between `before_all` with `before`.

That seems that we're trying to re-invent the shared contexts.

Let's remove the context-handling and simplify the hooks. That would be enough for the Discourse case, and we can discuss the extended API later.

danielwaterworth: @palkan, OK, done.

palkan: Thanks!

Issue #134, opened 2019-04-29
Add `let_it_be` aliases support


palkan:
Ref https://github.com/discourse/discourse/pull/7414#issuecomment-487507733

Add ability to register `let_it_be` aliases with the predefined options:

```ruby
TestProf::LetItBe.configure do |config|
  config.alias_to :fab!, refind: true
end
```

palkan:
Closed by #140

PR #136, opened 2019-04-30
Added let_it_be aliases

danielwaterworth:
#134

API:

    TestProf::LetItBe.configure do |config|
      config.alias_to :let_it_be_with_refind, refind: true
    end

    describe do
      let_it_be_with_refind(:foo) { Foo.create }

      # refind can still be overridden in the alias
      let_it_be_with_refind(:bar, refind: false) { Bar.create }
    end

### DIFF
--- a/lib/test_prof/recipes/rspec/let_it_be.rb
+++ b/lib/test_prof/recipes/rspec/let_it_be.rb
@@ -7,7 +7,21 @@ module TestProf
   # Just like `let`, but persist the result for the whole group.
   # NOTE: Experimental and magical, for more control use `before_all`.
   module LetItBe
+    class Configuration
+      def alias_to(name, **default_args)
+        LetItBe.define_let_it_be_alias(name, **default_args)
+      end
+    end
+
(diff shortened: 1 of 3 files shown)


lib/test_prof/recipes/rspec/let_it_be.rb
```diff
@@ -25,6 +39,12 @@ def modules
     # And we love cats!)
     PREFIX = RUBY_ENGINE == "jruby" ? "@__jruby_is_not_cat_friendly__" : "@😸"
 
+    def self.define_let_it_be_alias(name, **default_args)
+      define_method(name, &lambda do |identifier, **options, &blk|
```
palkan: Why creating `lambda`?

```ruby
define_method(name) do |identifier, **options, &blk|
  let_it_be(identifier, default_args.merge(options), &blk)
end
```

danielwaterworth: good catch

PR #157, opened 2019-08-25
Add column to show total time per factory call

tyleriguchi:
Make it easier to decide which factories to focus on for test performance.
Adds an additional column to the output for the simple printer to show how long each factory takes per call.

Screenshot of the changed output
Screen Shot 2019-08-25 at 1 59 28 PM

### DIFF
--- a/docs/factory_prof.md
+++ b/docs/factory_prof.md
@@ -12,15 +12,16 @@ Total top-level: 10286
 Total time: 299.5937s
 Total uniq factories: 119
 
- total   top-level   total time   top-level time            name
-  6091        2715    115.7671s         50.2517s            user
-  2142        2098     93.3152s         92.1915s            post
+ total   top-level   total time    time per call      top-level time            name
+  6091        2715    115.7671s          0.0426s            50.2517s            user
+  2142        2098     93.3152s          0.0444s            92.1915s            post
   ...
 ```
 
 It shows both the total number of the factory runs and the number of _top-level_ runs, i.e. not during another factory invocation (e.g. when using associations.)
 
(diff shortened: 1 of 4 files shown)


palkan:
I like the idea. Let’s fix the failing test and merge this.

tyleriguchi:
@palkan tests fixed!

palkan:
@tyleriguchi Thanks!
Could you please add a changelog entry and update the docs (example output https://test-prof.evilmartians.io/#/factory_prof)?

tyleriguchi:
Sure, I should be able to update that stuff today or later this week.

Also thanks for your work, between the gem and article talking about test perf, I was able to get our test suite from twenty minutes down to four. Was able to use factory prof to figure out which factories needed to be looked and fixed

docs/factory_prof.md
```diff
@@ -12,15 +12,16 @@ Total top-level: 10286
...
+ total   top-level   total time    time per call      top-level time            name
+  6091        2715    115.7671s          0.0426s            50.2517s            user
+  2142        2098     93.3152s          0.0444s            92.1915s            post
   ...
 ```
 
 It shows both the total number of the factory runs and the number of _top-level_ runs, i.e. not during another factory invocation (e.g. when using associations.)
 
 (**@since v0.9.0**) It also shows the time spent generating records with factories.
+(**@since v0.11.0**) It also shows the amount of time taken per factory call.
```
tyleriguchi: @palkan is this the correct way to note the next version?

palkan: Yep, looks good

palkan:
Thanks!

PR #168, opened 2020-01-07
Solve ruby 2.7 deprecation warnings

lostie:

To address the deprecation warnings when using ruby 2.7:

More info here: https://www.ruby-lang.org/en/news/2019/12/25/ruby-2-7-0-released/ 

- Replace the use of `Proc.new` with `block`
```shell
warning: Capturing the given block using Proc.new is deprecated; use `&block` instead
```

- Use double splat where necessary to align parameters definition with Ruby 2.7/3.x  

```shell
warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
```

- Rename Rubocop cop name.

⚠️ Currently the github `setup-ruby` action does not support ruby 2.7, so I was unable to add that version to the CI workflows, even though I got it tested successfully on my local machine.

More info:
- https://github.com/actions/setup-ruby
- https://github.com/actions/virtual-environments/blob/master/images/linux/Ubuntu1804-README.md

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -26,7 +26,7 @@ Style/TrailingCommaInArrayLiteral:
 Style/TrailingCommaInHashLiteral:
   EnforcedStyleForMultiline: no_comma
 
-Layout/AlignParameters:
+Layout/ParameterAlignment:
   EnforcedStyle: with_first_parameter
 
 # See https://github.com/rubocop-hq/rubocop/issues/4222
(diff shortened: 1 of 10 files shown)


lostie:
It seems https://github.com/eregon/use-ruby-action/ might be a viable alternative to be able to test it against more recent ruby versions.

palkan:
Thanks!

palkan:
> It seems https://github.com/eregon/use-ruby-action/ might be a viable alternative

Great idea. Would you like to propose a PR with this migration?

PR #169, opened 2020-01-07
Github CI workflow installs ruby with use-ruby-action

lostie:

The default github action to install ruby (`actions/setup-ruby@v1`) doesn't not have support for the most recent versions of ruby. The `use-ruby-action` (https://github.com/eregon/use-ruby-action) seems to be a good alternative for it.

Changed the github workflows to make use of `use-ruby-action` instead of  `setup-ruby` action to install ruby.

Also:
- Ensure the `Rubocop` version (which is a development dependency) is updated to be compliant with the format of the `.rubocop.yml`. Locally I was getting the following warning:

```
Warning: unrecognized cop Layout/ParameterAlignment found in .rubocop.yml
```

which was updated in https://github.com/palkan/test-prof/pull/168 since CI was failing the `Rubocop` check.

### DIFF
--- a/.github/workflows/docs-lint.yml
+++ b/.github/workflows/docs-lint.yml
@@ -15,10 +15,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v1
-    - name: Set up Ruby 2.6
-      uses: actions/setup-ruby@v1
+    - name: Set up Ruby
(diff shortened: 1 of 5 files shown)


palkan:
Nice 👍 

Can we also migrate the `rspec-jruby` workflow to use this action?

palkan:
Thank you!

PR #170, opened 2020-01-29
Improve `RSpec/AggregateExamples` cop

pirj:

Improve `RSpec/AggregateExamples` cop (formerly `RSpec/AggregateFailures`).
Name change is semantic, since what cop suggests is to aggregate examples, and the feature of RSpec that allows for doing this painlessly is known as Aggregate Failures, but it's not directly related to what the cop does.

Block expectation syntax is deliberately not supported due to:
 - `subject { -> { ... } }` syntax being hard to detect
 - aggregation should use composition with `.and`
 - aggregation of the `not_to` is barely possible when a matcher doesn't
 provide a negated variant
 - aggregation of block syntax with non-block syntax should be in a
 specific order

**Known caveats**:
The usages if `its` that are testing private methods/readers will result in spec failure. Test only public interface!

Matchers with side effects might affect subsequent expectations when aggregated. A warning is added to the offence.

- significantly improved `its` support;
- fixed `--auto-correct` showing all detected offenses as corrected;
- fixed offense detection of matchers with side effects with qualifiers;
- removed redundant spec coverage;
- divided the code into logical parts (more on that below), that allowed removing disabling of `Metrics/ClassLength` on the cop;
- extracted specs for `its` and matchers with side effects to dedicated spec files;
- ... and much more!

This cop was forked from `RSpec/AggregateFailures` and developed internally at @toptal (and part of that work was generously sponsored by them), but it might have diverged since then.
I've overwritten the existing cop, so please pay close attention to features that might have been added after forking, and happen to miss from this implementation.

The cop was originally submitted to RuboCop RSpec (https://github.com/rubocop-hq/rubocop-rspec/pull/726) but rejected due to its primary focus on spec performance, and also its unwieldiness.
There are some nice comments there.

This https://github.com/rubocop-hq/rubocop-rspec/pull/726#issuecomment-483458278 highlights some of the design decisions.

### Not Done

The cop seems to be more or less feature-complete and bug-free. It was tested and regularly used on a 1M+ LoC codebase for almost a year. All edge cases and feedback found its way in the cop.
However, some of the things might be surprising, like lack of auto-correction for the block expectation syntax, unsafe matchers, and impaired document formatter output due to removals of example description along correction.

1. For the latter, example descriptions, there is a solution used to extract expectations to matchers:
```ruby
def be_cool
  be_cold
    .and be_fresh
end
```
Native RSpec's composing capabilities will provide a proper description for failures and for example doc.

2. Not compounded by default
```ruby
is_expected.to be_cold
is_expected.to be_fresh
```
are not automatically aggregated to a compound expectation even though it could:
```
is_expected
  .to be_cold
  .and be_fresh
```
There's a space for improving this, or a dedicated cop (probably more generic and less performance-focused) could do that. The compoind notation allows extracting parts of expectation to methods without sacrificing self-describability.

3. It would be possible to extract UnsafeMatchers cop, but there are edge cases of using several cops cross-concerns that arise with that extraction.
Some thoughts on this topic https://github.com/rubocop-hq/rubocop-rspec/pull/726#discussion_r314961256 https://github.com/rubocop-hq/rubocop-rspec/pull/726#issuecomment-558826999

4. Found it in the last moment.

Add an example where a variable/method is an argument to `its`:
```ruby
its(something) { is_expected.to ... }
```
I'm not certain if all those cases can/should be auto-corrected to
```ruby
expect(subject.public_send(something)).to ...
```
This can be tackled in the future. Meanwhile added a spec to ignore this.

### Checklist

- [x] Leftover: remove unnecessary parts of `language.rb`
- [x] `all_expectations?` can be rewritten
- [ ] I've squashed commits


lib/test_prof/factory_bot.rb
```diff
@@ -5,7 +5,7 @@ module TestProf # :nodoc: all
 
   FACTORY_GIRL_NAMES.find do |name, cname|
     TestProf.require(name) do
-      TestProf::FactoryBot = Object.const_get(cname)
+      TestProf.const_set("FactoryBot", Object.const_get(cname))
```
pirj: Unrelated, the issue resolved itself in some other way. Will remove.

lib/test_prof/cops/rspec/aggregate_examples/its.rb
```diff
@@ -0,0 +1,91 @@
...
+        #
+        # @example `its` with multi-element array argument is ambiguous
+        #
+        #   # Does not support `its` with multi-element array argument due to
+        #   # an ambiguity. Transformation depends on the type of the subject:
+        #   # - a Hash: `hash[element1][element2]...`
+        #   # - and arbitrary type: `hash[element1, element2, ...]`
+        #   # It is impossible to infer the type to propose a proper correction.
+        #
+        #   its(['ambiguous', 'elements']) { ... }
```
pirj: This `dig` auto-correction (in action in the previous implementation) seems to be an incorrect correction for cases outside arrays and hashes.
```ruby
# Works for nested arrays and hashes
expect(subject.dig('ambiguous', 'elements')).to eq [100, 100]
```
```
"sdds"[1, 2] # => "dd"
"sdds".dig(1, 2) # NoMethodError: undefined method `dig' for "sdds":String
```
The following passes:
```
require 'rspec/its'
RSpec.describe do
  subject { 'qwer' }
  its([1,2]) { is_expected.to eq 'we' }
end
```

pirj:
I have no good idea of what happens to liche.

lib/test_prof/cops/rspec/aggregate_examples/its.rb
```diff
@@ -0,0 +1,91 @@
...
+          def transform_its(body, arguments)
+            argument = arguments.first
+            replacement = case argument.type
+                          when :array
+                            key = argument.values.first
+                            "expect(subject[#{key.source}])"
+                          else
+                            property = argument.value
+                            "expect(subject.#{property})"
+            end
```
pirj: RuboCop, are you even watching?

lib/test_prof/cops/rspec/aggregate_examples/its.rb
```diff
@@ -0,0 +1,91 @@
...
+          #     an ambiguity, when SUT can be a hash, and result will be defined
+          #     by calling `[]` on SUT subsequently, e.g. `subject[one][two]`,
+          #     or any other type of object implementing `[]`, and then all the
+          #     array arguments are passed to `[]`, e.g. `subject[one, two]`.
+          def_node_matcher :example_for_autocorrect?, <<-PATTERN
+            [ #super !#its_with_multiple_element_array_argument? ]
+          PATTERN
+
+          def_node_matcher :its_with_multiple_element_array_argument?,
+            <<-PATTERN
```
pirj: Weird indentation.

lib/test_prof/cops/rspec/aggregate_examples/matchers_with_side_effects.rb
```diff
@@ -0,0 +1,83 @@
...
+        #
+        #   # bad, but isn't automatically correctable
+        #   describe do
+        #     it { is_expected.to validate_presence_of(:comment) }
+        #     it { is_expected.to be_valid }
+        #   end
+        #
+        # @internal
+        #   Support for taking special care of the matchers that have side
+        #   effects, i.e. leave the subject in a modified stte.
```
pirj: s/stte/state/

Wondering why forspell didn't catch it.

lib/test_prof/cops/rspec/aggregate_examples/matchers_with_side_effects.rb
```diff
@@ -0,0 +1,83 @@
...
+          MSG_FOR_EXPECTATIONS_WITH_SIDE_EFFECTS =
+            "Aggregate with the example at line %d. IMPORTANT! Pay attention " \
+            "to the expectation order, some of the matchers have side effects."
+
+          private
+
+          def message_for(example, first_example)
+            return super unless example_with_side_effects?(example)
+
+            MSG_FOR_EXPECTATIONS_WITH_SIDE_EFFECTS % first_example.loc.line
```
pirj: Use `.format`.

lib/test_prof/cops/rspec/aggregate_examples.rb
```diff
@@ -0,0 +1,177 @@
...
+        # Clusters of examples that can be aggregated without losing any
+        # information (e.g. metadata or docstrings)
+        def example_clusters_for_autocorrect(example_node)
+          examples_in_group = example_node.parent.each_child_node(:block)
+            .select { |example| example_for_autocorrect?(example) }
+          example_clusters(examples_in_group)
+        end
+
+        def message_for(_example, first_example)
+          MSG % first_example.loc.line
```
pirj: Use `.format`.

spec/test_prof/cops/rspec/aggregate_examples/its_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+        end
+        its(['two']) { is_expected.to be(false) }
+      end
+    RUBY
+
+    it "detects and autocorrects" do
+      expect_offense(offensive_source)
+      expect_correction(good_source)
+    end
+  end
```
pirj: Add an example where a variable/method is an argument to `its`:
```ruby
its(something) { is_expected.to ... }
```
I'm not certain if all those cases can/should be auto-corrected to
```ruby
expect(subject.public_send(something)).to ...
```

spec/test_prof/cops/rspec/aggregate_examples_spec.rb
```diff
@@ -0,0 +1,620 @@
...
+        #{group} 'some docstring' do
+          it { is_expected.to be_awesome }
+          it { expect(subject).to be_amazing }
+          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Aggregate with the example at line 2.
+          it { expect(article).to be_brilliant }
+          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Aggregate with the example at line 2.
+        end
+      RUBY
+
+      good_source = <<-RUBY
```
pirj: Now when I look at it, I'm not sure if usage of `good_source`/`offensive_source` is worth it compared to:
```ruby
expect_offense(<<~RUBY)
  ...
RUBY

expect_correction(<<~RUBY)
  ...
RUBY
```

spec/test_prof/cops/rspec/aggregate_examples_spec.rb
```diff
@@ -0,0 +1,620 @@
...
+  it_behaves_like "detects and autocorrects in example group",
+    :example_group
+
+  # Non-expectation statements can have side effects, when e.g. being
+  # part of the setup of the example.
+  # Examples containing expectations wrapped in a method call, e.g.
+  # `expect_no_corrections` are not considered aggregatable.
+  context "with examples with non-expectation statements" do
+    it "does not detect offenses" do
+      expect_no_offenses(<<-RUBY)
```
pirj: Non-squiggly HEREDOC produces output that is hard to read due to excessive and inconsistent identation.

spec/test_prof/cops/rspec/aggregate_examples_spec.rb
```diff
@@ -0,0 +1,620 @@
...
+      end
+    RUBY
+
+    it "detects and autocorrects" do
+      expect_offense(offensive_source)
+      expect_correction(good_source)
+    end
+  end
+
+  context "with an expectation with chaining matchers" do
```
pirj: s/chaining/compound/

spec/test_prof/cops/rspec/aggregate_examples_spec.rb
```diff
@@ -0,0 +1,620 @@
...
+      end
+    RUBY
+
+    it "detects and autocorrects" do
+      expect_offense(offensive_source)
+      expect_correction(good_source)
+    end
+  end
+
+  # With examples have docstrings, it is incorrect to aggregate them, since
```
pirj: s/With/When/

spec/test_prof/cops/rspec/aggregate_examples_spec.rb
```diff
@@ -0,0 +1,620 @@
...
+  # either docstring is lost, either needs to be joined with the others,
```
pirj: s/either needs/either it needs/

spec/test_prof/cops/rspec/aggregate_examples_spec.rb
```diff
@@ -0,0 +1,620 @@
...
+      end
+    RUBY
+
+    it "detects an offense, but does not correct it" do
+      expect_offense(offensive_source)
+      expect_no_corrections
+    end
+  end
+
+  context "when all examples have names" do
```
pirj: s/names/docstrings/

spec/test_prof/cops/rspec/aggregate_examples_spec.rb
```diff
@@ -0,0 +1,620 @@
...
+
+    it "detects an offense, but does not correct it" do
+      expect_offense(offensive_source)
+      expect_no_corrections
+    end
+  end
+
+  context "with block expectation syntax" do
+    fair_source = <<-RUBY
+      describe '#complete' do
```
pirj: `'#complete'` does not add any value here.

spec/test_prof/cops/rspec/aggregate_examples_spec.rb
```diff
@@ -0,0 +1,620 @@
...
+
+    it "detects and autocorrects" do
+      expect_offense(offensive_source)
+      expect_correction(good_source)
+    end
+  end
+
+  context "with several examples separated by newlines" do
+    offensive_source = <<-RUBY
+      RSpec.describe do
```
pirj: `RSpec` is redundant starting from here.

lib/test_prof/cops/rspec/aggregate_failures.rb
```diff
@@ -1,170 +0,0 @@
...
-        GROUP_BLOCKS = %i[
-          describe context feature example_group
-        ].freeze
-
-        EXAMPLE_BLOCKS = %i[
-          it its specify example scenario
-        ].freeze
-
-        class << self
-          def supported?
```
pirj: Should I keep this check?

lib/test_prof/cops/rspec/aggregate_examples/its.rb
```diff
@@ -0,0 +1,91 @@
...
+        #   its('header', html: true) { is_expected.to include(text: 'hello') }
+        #
+        module Its
+          extend RuboCop::NodePattern::Macros
+
+          private
+
+          # It's impossible to aggregate `its` body as is, it needs to be
+          # converted to `expect(subject.something).to ...`
+          def new_body(node)
```
pirj: It's not too obvious (you can see `super` being used, and a mess of `include`/`prepend` in the cop base body) that some of the methods are overridden.
It makes sense to cross-reference methods that override each other.

spec/test_prof/cops/rspec/aggregate_examples/its_spec.rb
pirj: A prospect for a future improvement.

lib/test_prof/cops/rspec/aggregate_examples/its.rb
palkan: What about chaining the naming of the module? For example, `Aggregation` -> `BaseAggregation` (which indicates that this module should be included first). Similarly, `Its` -> `ItsAggregation` (hints that it extends `BaseAggregation`).

lib/test_prof/cops/rspec/aggregate_failures.rb
palkan: Yeah, let's keep it (and move to `test_prof/rubocop.rb`). We do not have restrictions in the gemspec, so this is the only way to handle the unsupported version (yeah, pretty old, but still could be used).

lib/test_prof/cops/rspec/aggregate_examples/its.rb
pirj: It turns out it's the way `EndAlignment` works for `case` and `begin`.

pirj:
Rearranged includes indicating their purposes.
Added back runtime RuboCop version check.
Moved default config injection code to `cops/inject`.

Kept `AggregateFailures` as an alias of `AggregateExamples` with a warning on usage. Tried to use RuboCop's obsoletion for that, but it's hard to inject there, and a bit scary, since it doesn't look like something that is expected to be played with and may be a subject to change.

Appreciate if you take another look @palkan

lib/test_prof/cops/rspec/aggregate_examples.rb
```diff
@@ -0,0 +1,199 @@
...
+      #     expect(number).to be_positive
+      #     expect(number).to be_odd
+      #   end
+      #
+      class AggregateExamples < Cop
+        include LineRangeHelpers
+        include MetadataHelpers
+        include NodeMatchers
+
+        # Methods from the following modules override and extend methods of this
```
palkan: 👍

lib/test_prof/cops/rspec/aggregate_failures.rb
```diff
@@ -1,168 +1,26 @@
...
-          match = body.arguments.first.source
-
-          if subject_attribute.array_type?
-            hash_keys = subject_attribute.values.map(&:value).join(", ")
-            attribute = "dig(#{hash_keys})"
-          else
-            attribute = subject_attribute.value
-          end
+      class AggregateFailures < AggregateExamples
+        raise "Remove me" if TestProf::VERSION >= "1.0"
```
palkan: 👍

pirj:
Squashed and green 👹 🙌 

BTW for the *not done*: I've omitted "Eat your own cat food" commit, since for some reason unless you specify `Enable: true` for the cop in `.rubocop.yml` explicitly, it's disabled. When I plug it into another project, it works without explicit enabling.
Also, just two minor offences were found ¯\_(ツ)_/¯

PR #177, opened 2020-02-11
Disable AggregateFailures by default

pirj:

Disable AggregateFailures by default and keep AggregateExamples enabled.

According to [docs, it had to be enabled explicitly](https://github.com/palkan/test-prof/blame/c01b309f4534e93b614dd2b79ac1f7f7ec1d92f3/docs/rubocop.md#L19) previously.

Disabled AggregateFailures in RuboCop default config.

AggregateExamples is enabled, so for those who have enabled AggregateFailures, double the offences will be shown, and a deprecation warning regarding AggregateFailures.
It will bring double the attention to deprecation, and fix is trivial - remove AggregateFailures from user RuboCop config.

### Checklist

- [-] I've added tests for this change
- [-] I've added a Changelog entry
- [-] I've updated a documentation

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -22,7 +22,7 @@ RSpec/AggregateExamples:
 # TODO: remove this one we hit 1.0
 RSpec/AggregateFailures:
   Description: Checks if example groups contain two or more aggregatable examples.
-  Enabled: true
+  Enabled: false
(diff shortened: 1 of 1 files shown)


palkan:
Thanks!

PR #178, opened 2020-02-28
Add state leakage detection to `let_it_be`

pirj:

State leakages happen if examples or hooks modify models.

From [`rspec-rails` docs](https://relishapp.com/rspec/rspec-rails/v/3-9/docs/transactions) on transactions and `before(:context)`:

> Even though database updates in each example will be rolled back, the object won't know about those rollbacks so the object and its backing data can easily get out of sync.

Most cases are caused by examples or the code under test modifies models that are shared between examples when not `reload`'ed or `refind`'ed.

Direct, or *first-order* state leakage is when the model passed to `let_it_be` is modified.
*Second-order* state leakage is when an associated model is modified.

State leakages are hard to track down to the root cause and lead to unpleasant surprises that usually accumulate error.
When the root cause changes its behavior, the problem becomes apparent with numerous failures.

In addition to the code under test that may modify models, there are other root causes, e.g.:

#### 😈 Nasty RSpec Hook

```ruby
let_it_be(:user) { create(:user) }

context 'when the user is approved' do
  before { user.update!(status: :approved) }

  it { is_expected.to be_allowed }
end

context 'on Sunday' do # open day!
  before { time_travel(:sunday) }

  it { is_expected.to be_allowed } # Haha, it's not because of Sunday, it's because the user is approved!
end
```

#### 😈 Nasty Factory Hook

It's not uncommon to use `after(:create)`/`after(:build)` to e.g. create some backreferences:
```ruby
FactoryBot.define do
  factory :contract do
    after(:create) do |contract|
      contract.customer.last_contract = contract
      contract.customer.save!
    end
  end
end
```
```ruby
let_it_be(:customer) { create(:customer) }

describe "with a contract" do
  let(:contract) { create(:contract, customer: customer) } # Looks pretty harmless, uh?

  # ...
end
```

I'm not going into details if it's a bad practice or not, especially since the same side effect of modifying the object served by `let_it_be` can be done via regular model hooks.
The fact is that `customer` is modified and in the subsequent example its `last_contract` will be filled in with a reference to a contract that doesn't exist. In some cases, it can be detected if one has proper FK constraints set, but typically this goes under the radar.

#### 💚 Fair Use

Some code under test inevitably mutates state.

```ruby
# code under test

def log_in(user)
  active_sessions << user.id
  user.update!(last_logged_in: Time.now)
end
```

```ruby
# test
let_it_be(:user) { create(:user) } # it will be made frozen

it { expect { log_in(user) }.to change(active_sessions).to include(user.id) # BOOM
```

A workaround here is to:

```diff
- let_it_be(:user) { create(:user) } # it will be made frozen
+ let_it_be(:user, freeze: false) { create(:user) } # it will NOT be made frozen
```
or even better:
```diff
- let_it_be(:user) { create(:user) } # it will be made frozen
+ let_it_be(:user, reload: true) { create(:user) }
+ # it will NOT be made frozen, but will be reloaded to avoid `last_logged_in` changes between examples
```

 - freeze models passed to `let_it_be` along with their associations
 - introduce an option to `let_it_be` to `freeze: false` (which is inferred from `reload: true` and `refind: true` as well)
 - introduce `let_it_be_defrost: true` RSpec metadata to skip leakage detection for the whole example group

Run against your projects.

The first-order detection is in production (ergh, test) use for more than a year and helped to detect and eliminate hundreds of cases of state leakage.
The second-order leakage detection revealed thousands of leakages.

**WARNING** This is a breaking change. To fix the code one of the following methods should be applied:

 - add reload: true/refind: true, typically it's significantly faster to reload the model than to re-create it from scratch before each example (two or even three orders of magnitude in some cases)
 - add freeze: false to those models that are being mutated in a way that is highly unlikely to affect other examples. USE WITH CAUTION - it's a time bomb, you don't know for sure when "highly unlikely" flips to "possibly"
- rewrite your Nasty Hooks and other problematic code

#### NOT Done

 - [-] FrozenError doesn't always point to where the frozen object was defined exactly. This may be improved [by referencing the object that was frozen](https://github.com/palkan/test-prof/pull/178#discussion_r387273143)

### Checklist

 - [x] I've updated the documentation

**NOTE:** This work was generously sponsored by @toptal. The first-order leakage was originally implemented by @jaimerson, and we hacked on second-order leakage detection with @AlexVKO.

### DIFF
--- a/lib/test_prof/recipes/rspec/let_it_be.rb
+++ b/lib/test_prof/recipes/rspec/let_it_be.rb
@@ -125,9 +125,6 @@ module TestProf
 
           # Prevent records that are marked as `freeze: false`, `reload: true`, and
           # `refind: true` from being frozen by walking the association tree.
-          # OPTIMIZE: is it the best place? can it be less tangled?
-          # TODO: check if it's not too late, that we freeze *after* this
-          # proc is getting called.
           Freezer.stoplist << record unless freeze
 
           instance_variable_set(:"#{TestProf::LetItBe::PREFIX}#{identifier}", record)
@@ -143,8 +140,8 @@ module TestProf
 
       def freeze(object)
         object.freeze
-        # OPTIMIZE: is this really needed? What if there is a circular reference?
-        # return object.each { |obj| freeze(obj) } if object.respond_to?(:each)
+        # Support `let_it_be` with a list of objects
+        object.each { |obj| freeze(obj) } if object.respond_to?(:each)
       end
 
       # Rerucsively freezes the object to detect modifications.
@@ -154,7 +151,7 @@ module TestProf
 
         record.freeze
 
-        # This is required to support `let_it_be` with `create_list`
+        # Support `let_it_be` with `create_list`
         return record.each { |rec| deep_freeze(rec) } if record.respond_to?(:each)
 
         # Freeze associations as well.
@@ -174,9 +171,8 @@ module TestProf
         end
       end
 
-      # Stoplist is needed to prevent freezing objects that are
-      # defined with `let_it_be`'s `reload: true`/`refind: true`
-      # options.
+      # Stoplist to prevent freezing objects that are defined with `let_it_be`'s
+      # `reload: true`/`refind: true`/`freeze: false` options during deep freezing.
       # NOTE: it's intentionally implemented as a thread-local var.
       def stoplist
         Thread.current[:"#{TestProf::LetItBe::PREFIX}stoplist"] ||= Set.new
--- a/lib/test_prof/recipes/rspec/let_it_be.rb
+++ b/lib/test_prof/recipes/rspec/let_it_be.rb
@@ -86,8 +86,8 @@ module TestProf
     end
 
     def let_it_be(identifier, **options, &block)
-      freeze = options.fetch(:freeze, !(options[:reload] || options[:refind])) &&
-        !metadata[:let_it_be_defrost]
+      freeze = metadata[:let_it_be_frost] &&
+        options.fetch(:freeze, !(options[:reload] || options[:refind]))
 
       initializer = build_let_it_be_initializer(identifier, freeze, &block)
       before_all(&initializer)
@@ -144,7 +144,7 @@ module TestProf
         object.each { |obj| freeze(obj) } if object.respond_to?(:each)
       end
 
-      # Rerucsively freezes the object to detect modifications.
+      # Rerucsively freezes the object to detect modifications
       def deep_freeze(record)
         return if record.frozen?
         return if stoplist.include?(record)
@@ -215,7 +215,8 @@ if defined?(::ActiveRecord::Base)
     end
 
     config.register_modifier :freeze do |record, val|
-      next record if val == false
+      # TODO: change this to `if val == false` when TestProf hits 1.0
+      next record unless val == true
 
       TestProf::LetItBe::Freezer.deep_freeze(record)
       record
@@ -223,7 +224,8 @@ if defined?(::ActiveRecord::Base)
   end
 else
   config.register_modifier :freeze do |record, val|
-    next record if val == false
+    # TODO: change this to `if val == false` when TestProf hits 1.0
+    next record unless val == true
 
     TestProf::LetItBe::Freezer.freeze(record)
     record
--- a/lib/test_prof/recipes/rspec/let_it_be.rb
+++ b/lib/test_prof/recipes/rspec/let_it_be.rb
@@ -75,7 +75,7 @@ module TestProf
     # And we love cats!)
     PREFIX = RUBY_ENGINE == "jruby" ? "@__jruby_is_not_cat_friendly__" : "@😸"
 
-    FROZEN_HASH_HINT = "\nIf you are using `let_it_be`, you may want to pass `reload: true` option to it."
+    FROZEN_ERROR_HINT = "\nIf you are using `let_it_be`, you may want to pass `reload: true` option to it."
 
     def self.define_let_it_be_alias(name, **default_args)
       define_method(name) do |identifier, **options, &blk|
@@ -146,7 +146,7 @@ module TestProf
 
           instance_variable_set(:"#{TestProf::LetItBe::PREFIX}#{identifier}", record)
         rescue => e
-          e.message << FROZEN_HASH_HINT if e.message.match?(/can't modify frozen Hash/)
+          e.message << FROZEN_ERROR_HINT if e.message.match?(/can't modify frozen/)
           raise e
         end
       end
@@ -227,8 +227,8 @@ end
 RSpec::Core::ExampleGroup.extend TestProf::LetItBe
 RSpec.configure do |config|
   config.after(:example) do |example|
-    if example.exception&.message&.match?(/can't modify frozen Hash/)
-      example.exception.message << TestProf::LetItBe::FROZEN_HASH_HINT
+    if example.exception&.message&.match?(/can't modify frozen/)
+      example.exception.message << TestProf::LetItBe::FROZEN_ERROR_HINT
     end
   end
 end
--- a/lib/test_prof/recipes/rspec/let_it_be.rb
+++ b/lib/test_prof/recipes/rspec/let_it_be.rb
@@ -124,12 +124,6 @@ module TestProf
         let_it_be_objects.each { |object| Freezer.deep_freeze(object) }
       end
 
-      prepend_after do |example|
-        if example.exception&.message&.match?(/can't modify frozen Hash/)
-          example.exception.message << FROZEN_HASH_HINT
-        end
-      end
-
       instance_variable_set(:"#{PREFIX}hooks_defined", true)
     end
 
@@ -231,3 +225,10 @@ if defined?(::ActiveRecord)
 end
 
 RSpec::Core::ExampleGroup.extend TestProf::LetItBe
+RSpec.configure do |config|
+  config.after(:example) do |example|
+    if example.exception&.message&.match?(/can't modify frozen Hash/)
+      example.exception.message << TestProf::LetItBe::FROZEN_HASH_HINT
+    end
+  end
+end
--- a/lib/test_prof/recipes/rspec/let_it_be.rb
+++ b/lib/test_prof/recipes/rspec/let_it_be.rb
@@ -139,8 +139,7 @@ module TestProf
             # But only if they are already loaded. If not yet loaded, they weren't
             # created by factories, and it's ok to mutate them.
 
-            # next unless record.association(reflection.to_sym).loaded?
-            next unless record.association_cached?(reflection.to_sym)
+            next unless record.association(reflection.to_sym).loaded?
 
             target = record.association(reflection.to_sym).target
             deep_freeze(target) if target.is_a?(::ActiveRecord::Base) || target.respond_to?(:each)
--- a/lib/test_prof/recipes/rspec/let_it_be.rb
+++ b/lib/test_prof/recipes/rspec/let_it_be.rb
@@ -157,9 +157,9 @@ module TestProf
     #
     # Stoplist holds two types of records, one is "stop", and another is "skip".
     # "skip" is to skip freezing objects that are defined with `let_it_be`'s
-    # `reload: true`/`refind: true`, but to proceed with deep freezing their
-    # associations. "stop" is for stop deep freezing for those records declared
-    # with `freeze: false`.
+    # `reload: true`, but to proceed with deep freezing their associations.
+    # "stop" is for stop deep freezing for those records declared with
+    # `freeze: false` and `refind: true`.
     module Stoplist
       class << self
         def skip?(record)
@@ -189,8 +189,9 @@ module TestProf
         end
       end
 
-      @skiplist = [] # Stack of example group-related variable definitions
-      @stoplist = [] # Stack of example group-related variable definitions
+      # Stack of example group-related variable definitions
+      @skiplist = []
+      @stoplist = []
     end
   end
 end
@@ -218,7 +219,7 @@ if defined?(::ActiveRecord::Base)
     config.register_modifier :refind do |record, val|
       next record unless val
 
-      TestProf::LetItBe::Stoplist.skip!(record)
+      TestProf::LetItBe::Stoplist.stop!(record)
 
       next record.refind if record.is_a?(::ActiveRecord::Base)
 
--- a/lib/test_prof/recipes/rspec/let_it_be.rb
+++ b/lib/test_prof/recipes/rspec/let_it_be.rb
@@ -38,8 +38,7 @@ module TestProf
       end
 
       def wrap_with_modifiers(mods, &block)
-        # return block if mods.empty?
-        mods = {freeze: true}.merge(mods)
+        return block if mods.empty?
 
         validate_modifiers! mods
 
@@ -184,9 +183,10 @@ if defined?(::ActiveRecord::Base)
   TestProf::LetItBe.configure do |config|
     config.register_modifier :reload do |record, val|
       next record unless val
-      next record.reload if record.is_a?(::ActiveRecord::Base)
 
       TestProf::LetItBe::Stoplist.push(record)
+
+      next record.reload if record.is_a?(::ActiveRecord::Base)
 
       if record.respond_to?(:map)
         next record.map do |rec|
@@ -198,9 +198,10 @@ if defined?(::ActiveRecord::Base)
 
     config.register_modifier :refind do |record, val|
       next record unless val
-      next record.refind if record.is_a?(::ActiveRecord::Base)
 
       TestProf::LetItBe::Stoplist.push(record)
+
+      next record.refind if record.is_a?(::ActiveRecord::Base)
 
       if record.respond_to?(:map)
         next record.map do |rec|
--- a/lib/test_prof/recipes/rspec/let_it_be.rb
+++ b/lib/test_prof/recipes/rspec/let_it_be.rb
@@ -175,22 +175,12 @@ module TestProf
         record.class.reflections.keys.each do |reflection|
           # But only if they are already loaded. If not yet loaded, they weren't
           # created by factories, and it's ok to mutate them.
-          next unless association_cached?(record, reflection.to_sym)
+          next unless record.association(reflection.to_sym).loaded?
 
           target = record.association(reflection.to_sym).target
           if target.is_a?(::ActiveRecord::Base) || target.is_a?(Array)
             deep_freeze(target, stoplist)
           end
-        end
-      end
-
-      if ActiveRecord::VERSION::MAJOR >= 5
-        def self.association_cached?(record, reflection)
-          record.association_cached?(reflection)
-        end
-      else
-        def self.association_cached?(record, reflection)
-          record.association_cache[reflection]
         end
       end
     end
--- a/lib/test_prof/recipes/rspec/let_it_be.rb
+++ b/lib/test_prof/recipes/rspec/let_it_be.rb
@@ -81,10 +81,34 @@ module TestProf
       end
     end
 
+    # Some of the examples might (unwillingly, or deliberately) update
+    # model attributes.
+    # Unwillingly - if the underlying code under test modifies models, e.g.
+    # modifies `updated_at` attribute.
+    # Deliberately - if models are updated in `before` hooks or examples
+    # themselves instead of creating models in a proper state initially.
+    #
+    # It doesn't really matter if the database is modified or not since
+    # it's rolled back to a pristine state.
+    # However, since models created with `let_it_be` are shared between
+    # the examples, non-reloaded changes to models remain and leak between
+    # examples.
+    #
+    # This leads to unpredictable failures, and in worst case scenario
+    # examples that implicitly depend on other examples.
+    #
+    # Root cause is hard to track down, especially with random example
+    # execution order. A spec might fail with --seed 1001, but pass with
+    # 1002 & 1003.
+    #
+    # With many shared models between many examples, it's also hard to
+    # track down the example and exact place in the code that modifies
+    # the model. Even though the fix is trivial - to add set `refind` or
+    # `reload` options, it's rarely obvious where it should be set
+    # exactly.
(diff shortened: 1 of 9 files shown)


pirj:
> docs/let_it_be.md:134: Automagic (suggestions: Automatic, Auto magic, Auto-magic)

I'm not sure whether to "fix" it or to add to the dictionary :D

lib/test_prof/recipes/rspec/let_it_be.rb
```diff
@@ -75,24 +75,23 @@ def validate_modifiers!(mods)
...
       end
     end
 
     def let_it_be(identifier, **options, &block)
-      initializer = proc do
-        instance_variable_set(:"#{PREFIX}#{identifier}", instance_exec(&block))
-      end
+      freeze = options.fetch(:freeze, !(options[:reload] || options[:refind]))
 
-      if within_before_all?
```
pirj: This isn't necessary - `before_all` also does that check.

lib/test_prof/recipes/rspec/let_it_be.rb
```diff
@@ -114,6 +113,79 @@ def define_let_it_be_methods(identifier, **modifiers)
...
+
+    # Exception needs to be handled both here and in `handle_frozen_hash_error`
+    # because if it is raised in before_all it isn't caught in `after` block and
+    # if it's inside the example it isn't raised so it has to be handled in `after`.
+    def build_let_it_be_initializer(identifier, freeze, &block)
+      proc do
+        begin
+          record = instance_exec(&block)
+          if freeze
+            # FIXME: When models and their associations are defined with different
```
pirj: This should better be fixed sooner than later.

lib/test_prof/recipes/rspec/let_it_be.rb
```diff
@@ -114,6 +113,79 @@ def define_let_it_be_methods(identifier, **modifiers)
...
+            # options, e.g. `reload: true` and without it, the ones that are not
+            # supposed to be frozen will still be frozen here. Add those with
+            # `reload: true`/`refind: true`/`freeze: false` to ignore list.
+            let_it_be_objects = instance_variable_get(:"#{TestProf::LetItBe::PREFIX}let_it_be_objects")
+            let_it_be_objects ||= instance_variable_set(:"#{TestProf::LetItBe::PREFIX}let_it_be_objects", [])
+            let_it_be_objects << record
+          end
+
+          instance_variable_set(:"#{TestProf::LetItBe::PREFIX}#{identifier}", record)
+        rescue => e
```
pirj: I've modified this to rescue from `FrozenError`, but it's not in Ruby 2.4.
Wondering how [this](https://github.com/palkan/test-prof/pull/178/files#diff-cb255553719186df25a3969e3f716618R18) doesn't fail:
```ruby
expect { ... }.to raise_error(FrozenError, /can't modify frozen Hash/)
```

lib/test_prof/recipes/rspec/let_it_be.rb
```diff
@@ -114,6 +113,89 @@ def define_let_it_be_methods(identifier, **modifiers)
...
+        record.freeze
+
+        return record.each { |rec| deep_freeze(rec) } if record.respond_to?(:each)
+
+        # Freeze associations as well.
+        #
+        # NOTE: `reload` statements in test or production code will cause
+        # a `FrozenError`. In case the use of `reload` cannot be avoided, use
+        # `reload: true` in `let_it_be` declaration.
+        return unless record.is_a?(::ActiveRecord::Base)
```
palkan: We need to check for `defined?(ActiveRecord::Base)` first. Not everyone uses ActiveRecord)

lib/test_prof/recipes/rspec/let_it_be.rb
```diff
@@ -114,6 +113,89 @@ def define_let_it_be_methods(identifier, **modifiers)
...
+      # Prevent hooks from being defined several times
+      return if instance_variable_get(:"#{PREFIX}hooks_defined")
+
+      before(:all) do
+        let_it_be_objects = instance_variable_get(:"#{TestProf::LetItBe::PREFIX}let_it_be_objects")
+
+        let_it_be_objects.each { |object| Freezer.deep_freeze(object) }
+      end
+
+      prepend_after do |example|
```
palkan: Can we make this hook global? (`config.after`) Probably, with checking for the presence of `let_it_be_objects` inside it.

lib/test_prof/recipes/rspec/let_it_be.rb
```diff
@@ -114,6 +113,89 @@ def define_let_it_be_methods(identifier, **modifiers)
...
+        if example.exception&.message&.match?(/can't modify frozen Hash/)
```
palkan: Why Hash, btw?

What if we have:

```ruby
let_it_be(:users) { create_pair(:user) }

context "with one more user" do
  before { users << create(:user) }
end
```

(The example is very synthetic but anyway)

palkan: We can also take a look at `exception.receiver`. For example, we can track all the let_it_be-frozen objects (e.g., by setting a _special_ instance variable or by using a WeakMap).

lib/test_prof/recipes/rspec/let_it_be.rb
```diff
@@ -114,6 +113,89 @@ def define_let_it_be_methods(identifier, **modifiers)
...
+        #
+        # NOTE: `reload` statements in test or production code will cause
+        # a `FrozenError`. In case the use of `reload` cannot be avoided, use
+        # `reload: true` in `let_it_be` declaration.
+        return unless record.is_a?(::ActiveRecord::Base)
+
+        record.class.reflections.keys.each do |reflection|
+          # But only if they are already loaded. If not yet loaded, they weren't
+          # created by factories, and it's ok to mutate them.
+          next unless association_cached?(record, reflection.to_sym)
```
palkan: Doesn't `record.association(reflection.to_sym).loaded?` work in all Rails 4+?

lib/test_prof/recipes/rspec/let_it_be.rb
pirj: Good idea, I'll give it a shot. At a glance it seems like both hooks can be made global.

lib/test_prof/recipes/rspec/let_it_be.rb
pirj: Fair point on `create_pair`. `let_it_be` had no `reload` support for arrays when I touched leakage detection about a year ago, so it was always a `Hash`. I'll do a quick fix.

pirj: > we can track all the let_it_be-frozen objects

That would make a huge difference to have a lead on the ~suspect~ victim at all times.

lib/test_prof/recipes/rspec/let_it_be.rb
pirj: Makes sense. I'll add a guard statement before this one to check if `AR::B` is defined. Everything below this relates to AR.

lib/test_prof/recipes/rspec/let_it_be.rb
pirj: Can't tell this off the top of my head, will check. 👍

lib/test_prof/recipes/rspec/let_it_be.rb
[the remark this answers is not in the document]
palkan: Oh, it has been [added only in Ruby 2.7](https://rubyreferences.github.io/rubychanges/2.7.html#frozenerror-receiver-argument). You can check it by running:

```sh
ruby -e 'p (begin; [0].freeze << 1; rescue FrozenError => e; e.receiver; end)'
```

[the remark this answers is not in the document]
pirj: Loving it 👍

lib/test_prof/recipes/rspec/let_it_be.rb
```diff
@@ -114,6 +113,53 @@ def define_let_it_be_methods(identifier, **modifiers)
...
+        record.freeze
+
+        return record.each { |rec| deep_freeze(rec, stoplist) } if record.respond_to?(:each)
+
+        # Freeze associations as well.
+        #
+        # NOTE: `reload` statements in test or production code will cause
+        # a `FrozenError`. In case the use of `reload` cannot be avoided, use
+        # `reload: true` in `let_it_be` declaration.
+        return unless defined?(::ActiveRecord)
```
palkan: Let's make it more specific (some gems could define `ActiveRecord` module without ActiveRecord itself to load their extensions):

```suggestion
        return unless defined?(::ActiveRecord::Base)
```

lib/test_prof/recipes/rspec/let_it_be.rb
```diff
@@ -38,14 +38,15 @@ def modifiers
...
+        # return block if mods.empty?
+        mods = {freeze: true}.merge(mods)
 
         validate_modifiers! mods
 
         -> {
           record = instance_eval(&block)
           mods.inject(record) do |rec, (k, v)|
-            LetItBe.modifiers.fetch(k).call(rec, v)
+            LetItBe.modifiers.fetch(k).call(rec, v, mods)
```
palkan: Let's not allow modifiers to know about each other.

We can add `frozen?` checks to `reload` and `refind` modifiers instead. In case both `reload: true` and `freeze: true` are present in this order, it won't hurt (and there is a theoretical situations when it could be useful: for example, if you change the record state from the _outside_, smth like `context; let_it_be {}; context; before_all { change_underlying_db_row }; end; end;`).

lib/test_prof/recipes/rspec/let_it_be.rb
```diff
@@ -145,7 +191,22 @@ def define_let_it_be_methods(identifier, **modifiers)
       end
       record
     end
+
+    config.register_modifier :freeze do |record, val, mods|
+      next record if val == false
```
palkan: Shouldn't we add the record to the `stoplist` here?

And make `stoplist` a _global registry_ of `let_it_be` objects. How it could be implemented? I'm still thinking about WeakMap.

That could solve the edge cases you mentioned in https://github.com/palkan/test-prof/pull/178#pullrequestreview-367996478.

palkan:
@pirj How are you doing? It seems that ball is on your side)

pirj:
It is, I'll address the review ASAP.

lib/test_prof/recipes/rspec/let_it_be.rb
pirj: > We can add frozen? checks to reload and refind modifiers instead.

The problem is not that they will attempt reloading a frozen model, but that the model that will be reloaded were made frozen. So the stoplist is to keep track of those with `reload: true`/`refind: true`.

> theoretical situations when it could be useful

But how it will be reloaded if it's frozen, this will cause an `FrozenError`, and there's no way to unfreeze it.

lib/test_prof/recipes/rspec/let_it_be.rb
pirj: It seems that it could be possible to detect the moment where we open the outermost transaction, i.e. where `within_before_all?` flips from `false` to `true`. Then it makes it possible to reset the set of `let_it_be`d objects when the older entries don't apply anymore.

That would allow reducing the lifecycle of this set from suite to one spec (or even an example group if this example group does not have common `let_it_be`/`before_all` with other example groups).

spec/integrations/fixtures/rspec/let_it_be_fixture.rb
```diff
@@ -31,7 +31,7 @@
   config.alias_to :let_with_refind, refind: true
 end
```
pirj: Out of curiosity, do we run those ` spec/integrations/` on CI?
`fd fix spec/integrations/fixtures/rspec | xargs -L1 rspec` failed for me on several examples:
```
  4) Something nested context is expected to eq true
     Failure/Error: specify { is_expected.to eq true }

       expected: true
            got: nil

       (compared using ==)
rspec ./spec/integrations/fixtures/rspec/rspec_stamp_fixture_tmpl.rb:12 # Something fail me
rspec ./spec/integrations/fixtures/rspec/rspec_stamp_fixture_tmpl.rb:24 # Something
    you
    can
    not
    patch me

rspec ./spec/integrations/fixtures/rspec/rspec_stamp_fixture_tmpl.rb:16 # Something fail me with tag
rspec ./spec/integrations/fixtures/rspec/rspec_stamp_fixture_tmpl.rb:35 # Something nested context is expected to eq true
```

```
  1) before_all + Isolator fails when within transaction
     Failure/Error: FailingJob.perform_async(false)

     Isolator::BackgroundJobError:
       You are trying to enqueue background job inside db transaction. In case of transaction failure, this may lead to data inconsistency and unexpected bugs
       Details: FailingJob ([false])
     # /Users/pirj/.rvm/gems/ruby-2.5.1@test-prof/gems/sidekiq-6.0.5/lib/sidekiq/client.rb:74:in `push'
     # /Users/pirj/.rvm/gems/ruby-2.5.1@test-prof/gems/sidekiq-6.0.5/lib/sidekiq/worker.rb:243:in `client_push'
     # /Users/pirj/.rvm/gems/ruby-2.5.1@test-prof/gems/sidekiq-6.0.5/lib/sidekiq/worker.rb:200:in `perform_async'
     # ./spec/integrations/fixtures/rspec/before_all_isolator_fixture.rb:65:in `block (3 levels) in <top (required)>'
     # /Users/pirj/.rvm/gems/ruby-2.5.1@test-prof/gems/activerecord-6.0.2.1/lib/active_record/connection_adapters/abstract/database_statements.rb:281:in `block in transaction'

Finished in 0.04757 seconds (files took 1.65 seconds to load)
3 examples, 1 failure

Failed examples:

rspec ./spec/integrations/fixtures/rspec/before_all_isolator_fixture.rb:62 # before_all + Isolator fails when within transaction
```

```
  1) User fabrication creates and reloads user
     Failure/Error: expect(User.find(user.id).name).to eq "John 1"

       expected: "John 1"
            got: "John 0"

       (compared using ==)
     # ./spec/integrations/fixtures/rspec/factory_doctor_fixture.rb:39:in `block (3 levels) in <top (required)>'

Finished in 0.11925 seconds (files took 1.52 seconds to load)
7 examples, 1 failure

Failed examples:

rspec ./spec/integrations/fixtures/rspec/factory_doctor_fixture.rb:37 # User fabrication creates and reloads user
```

(and segfautled a couple of times, too).
Ruby 2.5.1
isolator (0.6.1) in Gemfile.lock

lib/test_prof/recipes/rspec/let_it_be.rb
```diff
@@ -114,16 +115,77 @@ def define_let_it_be_methods(identifier, **modifiers)
...
+        end
+      end
+    end
+
+    # Stoplist to prevent freezing objects that are defined with `let_it_be`'s
+    # `reload: true`/`refind: true`/`freeze: false` options during deep freezing.
+    # To only keep track of objects that are available in current example group,
+    # `begin` adds a new layer, and `rollback` removes a layer of unrelated objects
+    # along with rolling back the transaction where they were created.
+    module Stoplist
```
pirj: Sorry, I wasn't sold on the idea of using `WeakRef`. Do you think this is straingtforward enough?

lib/test_prof/recipes/rspec/let_it_be.rb
```diff
@@ -114,16 +115,77 @@ def define_let_it_be_methods(identifier, **modifiers)
...
+        def include?(record)
+          @stoplist.any? { |layer| layer.include?(record) }
+        end
+
+        def push(record)
+          @stoplist.last.push(record)
+        end
+
+        def begin
+          @stoplist.push([])
```
pirj: It was tempting to use `Set.new` here, but:
```ruby
    x = Set.new
    a = []
    x.include? a # => true
    a << 1
    x.include? a # => false

    x = []
    a = []
    x.include? a # => true
    a << 1
    x.include? a # => true
```
And there's no `rehash` on Set

lib/test_prof/recipes/rspec/let_it_be.rb
```diff
@@ -114,16 +115,77 @@ def define_let_it_be_methods(identifier, **modifiers)
...
+        end
+      end
+
+      @stoplist = [] # Stack of example group-related variable definitions
+    end
   end
 end
 
-if defined?(::ActiveRecord)
+if defined?(::ActiveRecord::Base)
```
pirj: As you've mentioned above, having `ActiveRecord` in place is not a sign that it's used.

lib/test_prof/recipes/rspec/let_it_be.rb
```diff
@@ -145,7 +210,35 @@ def define_let_it_be_methods(identifier, **modifiers)
...
+      # TODO: change this to `if val == false` when TestProf hits 1.0
+      unless val == true
+        TestProf::LetItBe::Stoplist.push(record)
+        next record
+      end
+
+      TestProf::LetItBe::Freezer.deep_freeze(record)
+      record
+    end
   end
```
pirj: I've added a non-AR version here, like (simplified):
```ruby
else
  config.register_modifier :freeze do |record, val|
    record.freeze if val
  end
end
```
But removed it. We may consider adding it later on.

lib/test_prof/recipes/rspec/let_it_be.rb
```diff
@@ -145,7 +210,35 @@ def define_let_it_be_methods(identifier, **modifiers)
...
+end
+
+TestProf::BeforeAll.configure do |config|
+  config.before(:begin) do
+    TestProf::LetItBe::Stoplist.begin
+  end
+
+  config.after(:rollback) do
+    TestProf::LetItBe::Stoplist.rollback
+  end
```
pirj: I couldn't find a hook that would run on the outermost `begin` (where the first `before_all`'s block is called), or have a context to call `within_before_all?` for the sake of resetting the stoplist.
That made me rework the stoplist as a layer per example group structure, and use those two hooks to add and remove layers.

pirj:
Just in case, this is ready for review @palkan.

palkan:
> Just in case, this is ready for review

I'll take a look in a week or two (depending on how my quarantine will be going 😷).

pirj:
Take care! 💚 No rush.

lib/test_prof/recipes/rspec/let_it_be.rb
```diff
@@ -136,6 +198,9 @@ def define_let_it_be_methods(identifier, **modifiers)
 
     config.register_modifier :refind do |record, val|
       next record unless val
+
+      TestProf::LetItBe::Stoplist.push(record)
```
pirj: After giving it another thought, when `refind`/`reload` are in action, this doesn't mean that it's protected from second-order leakages. So it makes sense not to freeze itself, but still walk and freeze its associated records.

pirj:
We had this conversation in the comments, can't quickly find it. I've stumbled across an interesting behavior:
```
[1] pry(TestProf::LetItBe::Freezer)> record.association(:user).loaded?
=> false
[2] pry(TestProf::LetItBe::Freezer)> record.association_cached?(:user)
=> true
```

Another interesting piece (it might be getting late over here :)
```
[5] pry(TestProf::LetItBe::Freezer)> record.user
D, [2020-03-25T21:42:14.148186 #97407] DEBUG -- :   User Load (2.5ms)  SELECT "users".* FROM "users" WHERE "users"."id" = ? LIMIT ?  [["id", 1], ["LIMIT", 1]]
=> #<User:0x00007fe948a6aed0 id: 1, name: "Original Name", tag: nil>
```
I wouldn't expect a query to the database for a model created by the factory 😕 

Aggressive (excessive?) `refind` seems to be the culprit. Since associations are reset and fetched on-demand, they are not frozen anymore and vulnerable to modification.

The following:
```ruby
  describe "second order state leakage with `refind: true`", order: :defined do
    let_it_be(:post, refind: true) { create(:post, user: create(:user, name: "Original Name")) }

    it "leaks" do
      post.user.update!(name: "John Doe")
    end

    it { expect(post.user.name).to eq("Original Name") }
  end
```
fails, but not where you would expect (`update!`):
```
  1) Modification detection second order state leakage with `refind: true` is expected to eq "Original Name"
     Failure/Error: it { expect(post.user.name).to eq("Original Name") }

       expected: "Original Name"
            got: "John Doe"

       (compared using ==)
 ```
💥

palkan:
@pirj Could you please create a separate PR with the failing test?

pirj:
@palkan Sure, here you go https://github.com/palkan/test-prof/pull/185

spec/integrations/fixtures/rspec/let_it_be_fixture.rb
pirj: I've figured that out, there are deliberate failures in spec/integration/fixtures, and specs in spec/integration take care of checking the output.

lib/test_prof/recipes/rspec/let_it_be.rb
pirj: Let me reconsider this in the light of my recent discovery. I might have to simplify a couple of things.

pirj:
Looks green.
I'd love to run this on a large codebase with a lot of `let_it_be`s in it, but my current project doesn't yet use it, so I'm mostly doing this blindfolded.
Eager to see some early feedback from real projects.

pirj:
I guess it's ready for another critical look.

palkan:
> I guess it's ready for another critical look.

Sure. That's on my list. But I have a couple of bit releases in front of it 🤷🏻‍♂️

pirj:
Ping.
Given some time to think about this feature, I have had no new ideas.

pirj:
Traditional monthly ping.

palkan:
@pirj Thanks!

I reserved this Friday to work on this PR (yeah, finally 😀).

I have the following plan:
- I create a separate branch for let_it_be enhancements.
- Merge this PR into it.
- Try to use this branch in a couple of projects, maybe, do some minor refactoring.
- Implement features discussed here: https://github.com/palkan/test-prof/pull/178#pullrequestreview-367569092

pirj:
Awesome, thanks! 🎉

AlexVKO:
Wowwwww Nice!!!!

PR #184, opened 2020-03-16
Allow StackProf interval to be configured by ENV VAR

LynxEyes:

This PR adds the ability to configure stackprof's sample interval through an env var, similar to the rest of the options already do.

If you have a large test suite, using the default interval will create a massive dump file that is likely to end up being useless.
With this you can now configure the interval like so:

```
TEST_STACK_PROF=1 TEST_STACK_PROF_INTERVAL=10000 rspec
```

I added parsing for the env var while ensuring no behaviour changes for those not using the env var.
Also added a couple of line on the docs.

I haven't seen tests around this "get the init from env vars" anywhere - I may have missed them...
If they don't exist, should I create them?
If they do, my bad.. let me know what should I do then.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@
 
   See more in [#181](https://github.com/palkan/test-prof/issues/181).
 
+- Adds the ability to define stackprof's interval sampling by using `TEST_STACK_PROF_INTERVAL` env variable ([@LynxEyes][])
+
+  Now you can use `$ TEST_STACK_PROF=1 TEST_STACK_PROF_INTERVAL=10000 rspec` to define a custom interval (in microseconds).
+
 ## 0.11.3 (2020-02-11)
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,10 @@
 
   See more in [#181](https://github.com/palkan/test-prof/issues/181).
 
+- Adds the ability to define stackprof's interval sampling by using `TEST_STACK_PROF_INTERVAL` env variable ([@LynxEyes][])
```
palkan: Please, add your nickname to the end of the file in order for link anchor to work)

LynxEyes: I'm confused, wasn't that what I did? did I do it wrong?

LynxEyes: AH

palkan:
Thanks!

PR #185, opened 2020-03-26
Fix DB not being rolled back between sibling RSpec examples for `before_all`/`let_it_be`

pirj:
## The Problem

```ruby
describe "does not leak", order: :defined do
  include TestProf::FactoryBot::Syntax::Methods

  let_it_be(:user, refind: true) { create(:user, name: "Original Name") }

  it "leaks" do
    user.update!(name: "John Doe")
  end

  it "is not affected by the leak" do
    expect(user.name).to eq("Original Name")
  end
end
```

```
Modification detection
  association state leakage
    leaks
    suffers (FAILED - 1)

Failures:

  1) Modification detection association state leakage suffers
     Failure/Error: expect(post.user.name).to eq("Original Name")

       expected: "Original Name"
            got: "John Doe"
```

## Research

It surprised me that the database was not rolled back between sibling examples. Turns out `before(:all)` is not sufficient.

## The Solution

An additional nested transaction needs to be open/rolled back around every example.

### DIFF
--- a/lib/test_prof/before_all.rb
+++ b/lib/test_prof/before_all.rb
@@ -16,24 +16,26 @@ module TestProf
       attr_accessor :adapter
 
       def begin_transaction
+        raise "begin_transaction does not support a block" if block_given?
         raise AdapterMissing if adapter.nil?
 
         config.run_hooks(:begin) do
           adapter.begin_transaction
         end
-        yield
-      end
-
-      def within_transaction
-        yield
       end
 
       def rollback_transaction
-        raise AdapterMissing if adapter.nil?
-
         config.run_hooks(:rollback) do
           adapter.rollback_transaction
         end
+      end
+
+      def begin_example_transaction
+        adapter.begin_transaction
+      end
+
+      def rollback_example_transaction
+        adapter.rollback_transaction
       end
 
       def config
--- a/lib/test_prof/recipes/rspec/before_all.rb
+++ b/lib/test_prof/recipes/rspec/before_all.rb
@@ -9,26 +9,25 @@ module TestProf
       def before_all(&block)
         raise ArgumentError, "Block is required!" unless block_given?
 
-        return within_before_all(&block) if within_before_all?
+        return before(:all, &block) if within_before_all?
 
         @__before_all_activated__ = true
 
         before(:all) do
-          BeforeAll.begin_transaction do
-            instance_eval(&block)
-          end
+          BeforeAll.begin_transaction
+          instance_eval(&block)
         end
 
         after(:all) do
           BeforeAll.rollback_transaction
         end
-      end
 
-      def within_before_all(&block)
-        before(:all) do
-          BeforeAll.within_transaction do
-            instance_eval(&block)
-          end
+        before(:example) do
+          BeforeAll.begin_example_transaction
+        end
+
+        after(:example) do
+          BeforeAll.rollback_example_transaction
         end
       end
 
--- a/lib/test_prof/recipes/rspec/let_it_be.rb
+++ b/lib/test_prof/recipes/rspec/let_it_be.rb
@@ -86,11 +86,7 @@ module TestProf
         instance_variable_set(:"#{PREFIX}#{identifier}", instance_exec(&block))
       end
 
-      if within_before_all?
-        within_before_all(&initializer)
-      else
-        before_all(&initializer)
-      end
+      before_all(&initializer)
 
       define_let_it_be_methods(identifier, **options)
     end
--- a/spec/integrations/fixtures/rspec/before_all_custom_adapter_fixture.rb
+++ b/spec/integrations/fixtures/rspec/before_all_custom_adapter_fixture.rb
@@ -5,13 +5,31 @@ require_relative "../../../support/ar_models"
 require "test_prof/recipes/rspec/before_all"
 
 module MyAdapter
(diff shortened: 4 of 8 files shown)


lib/test_prof/before_all.rb
```diff
@@ -16,26 +16,28 @@ class << self
       attr_accessor :adapter
 
       def begin_transaction
+        raise "begin_transaction does not support a block" if block_given?
```
pirj, line 19: I have no strong preference here, but using a block seems to be redundant here.

lib/test_prof/before_all.rb
```diff
@@ -16,26 +16,28 @@ class << self
...
         end
-        yield
-      end
-
-      def within_transaction
-        yield
       end
 
       def rollback_transaction
-        raise AdapterMissing if adapter.nil?
```
pirj, line 32: Chances of rollback method to be called with `adapter.nil?` are nearing zero, considering there's already such a check in `begin_transaction`.

lib/test_prof/before_all.rb
```diff
@@ -16,26 +16,28 @@ class << self
...
-
         config.run_hooks(:rollback) do
           adapter.rollback_transaction
         end
       end
 
+      def begin_example_transaction
```
pirj, line 33: Those two methods are used to open/rollback a transaction around examples. The existing ones could have been reused, but this way it seems to be more clear, and no conditional is needed to decide whether to run hooks or not.

lib/test_prof/recipes/rspec/let_it_be.rb
```diff
@@ -86,11 +86,7 @@ def let_it_be(identifier, **options, &block)
         instance_variable_set(:"#{PREFIX}#{identifier}", instance_exec(&block))
       end
 
-      if within_before_all?
```
pirj, line 89: There's this check already in `before_all` itself.

lib/test_prof/before_all.rb
```diff
@@ -16,26 +16,28 @@ class << self
...
+        raise "begin_transaction does not support a block" if block_given?
         raise AdapterMissing if adapter.nil?
 
         config.run_hooks(:begin) do
           adapter.begin_transaction
         end
-        yield
-      end
-
-      def within_transaction
```
pirj, line 27: Direct yield. I imagine it was used to indicate that something is done in a transaction, but that was actually misleading.

lib/test_prof/recipes/rspec/before_all.rb
```diff
@@ -9,26 +9,25 @@ module RSpec
...
+          BeforeAll.begin_transaction
+          instance_eval(&block)
         end
 
         after(:all) do
           BeforeAll.rollback_transaction
         end
-      end
 
-      def within_before_all(&block)
```
pirj, line 27: This boils down to `before(:all, &block)` since `within_transaction` was a passthrough. I've inlined the only usage.

spec/integrations/fixtures/rspec/before_all_custom_adapter_fixture.rb
```diff
@@ -5,13 +5,31 @@
 require "test_prof/recipes/rspec/before_all"
 
 module MyAdapter
+  @nested_transaction_count = 0
```
pirj, line 8: We didn't use nested transactions if we didn't have nested example groups.
Now one transaction is open for an example group, and another for the example itself, so the test adapter needs to be capable of providing SAVEPOINT support to avoid the error when nested `BEGIN` is called.

spec/integrations/fixtures/rspec/before_all_isolator_fixture.rb
```diff
@@ -38,20 +38,22 @@ class User < ActiveRecord::Base
 end
 
 describe "before_all + Isolator", :transactional do
-  before_all do
-    @user = User.create
-    SampleJob.perform_async(true)
-    @user.commited = true
-  end
+  context "with an implicit transaction" do
```
pirj, line 41: Not 100% certain if the intention for this test was to demonstrate that Isolator blows up **after** the example itself.

spec/integrations/fixtures/rspec/let_it_be_fixture.rb
```diff
@@ -190,3 +190,17 @@
     end
   end
 end
+
+describe "does not leak", order: :defined do
```
pirj, line 194: That scenario that helped to discover the issue. Slightly simplified compared to the original.

pirj:
@palkan Hope you're ok.

Please take a look when you have a moment. I got to the bottom of the problem and fixed it.
I reckon this should be better merged before #178

palkan:
This is the responsibility of `config.use_transactional_tests = true` or similar (e.g., database_cleaner), not `before_all`. Let's not re-invent the wheel.

I think, adding a note to the documentation saying that transactional examples must be enabled to avoid state leak.

pirj:
Agree, will add 👍

Should I add a test to indicate that behaviour?
Maybe we should issue a warning, or even fail with a message to turn on transactional tests if this option is turned off?

palkan:
> Maybe we should issue a warning, or even fail with a message to turn on transactional tests if this option is turned off?

Ideally, we should. But that's not easy to detect. There could be different way of implementing transactions examples, not only the default one. Even custom hooks. So, I think, notice in the documentation is enough for now.

PR #187, opened 2020-04-19
Use rspec seed to make SAMPLE consistent

stefkin:

It makes sample of test cases consistent with seed.

It should help to reproduce issues which may appear during SAMPLE run.

Also helps in a sutation where you want to measure sample group iteratively. 

I create a random number generator using rspec's seed and use it in Array#sample call

I couldn't figure out how to test this for minitest (it doesn't have example specific output), suggestions are welcome! 

Please check out my tests. Are 10 runs ok to make sure that seed works consistently?

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -549,3 +549,4 @@ Fixes [#10](https://github.com/palkan/test-prof/issues/10).
 [@lostie]: https://github.com/lostie
 [@pirj]: https://github.com/pirj
 [@LynxEyes]: https://github.com/LynxEyes
+[@stefkin]: https://github.com/stefkin
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+- SAMPLE and SAMPLE_GROUP work consistently with seed in RSpec. ([@stefkin][])
+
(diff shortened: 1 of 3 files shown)


stefkin:
I figured out how to test minitest in a similar way (using -v option).

But in order to get minitest's seed I'd need to hook into plugin_test_prof.

Would it be ok to move TestProf::MinitestSample#run logic into the plugin?

palkan:
> But in order to get minitest's seed I'd need to hook into plugin_test_prof. Would it be ok to move TestProf::MinitestSample#run logic into the plugin?

Can we leave the sampling logic where it is and add a `TestProf::MinitestSample.random_generator` instead which we can populate from the plugin? Will it work?

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+- SAMPLE and SAMPLE_GROUP work consistently with seed in RSpec. ([@stefkin][])
```
palkan, line 5: Please, add a shortcut with the URL to the bottom of the file. Currently, it's not clickable: https://github.com/palkan/test-prof/blob/903649336e49ec042aa85d6a8530499396ae14a0/CHANGELOG.md

stefkin: done, thanks

PR #190, opened 2020-05-26
Fix documentation link

sallesma:
Just a typo or a link that has moved :)


sallesma:
Not sure to understand how this change can break the build. Was it already broken?

palkan:
@pirj Could you please take a look? Probably, new RuboCop release broke something with AggregateFailures cop.

pirj:
@sallesma Upstream changes ¯\\\_(ツ)_/¯
Fixed in https://github.com/palkan/test-prof/pull/191

pirj:
Thanks for the fix @sallesma !

sallesma:
That was so much work! 😉 

Thanks for the repo, my life is about to get better

PR #198, opened 2020-08-06
Fix let_it_be with Rails3 when outside of transactions

### LINKED ISSUE #197: let_it_be leads to ActiveRecord invalid savepoint issues in Rails3
### Tell us about your environment

**Ruby Version:**

`2.3.4`

**Framework Version (RSpec, Minitest, FactoryGirl, Rails, whatever):**

`Rspec 3.7.0`
`Rails 3.2.22.5`
`Mysql 5.7`

**TestProf Version:**

`0.7.5`

### What did you do?

Using `let_it_be` within the specs in the following way: `let_it_be(:user) { create(:user) }`

### What did you expect to happen?

I expected `let_it_be` to correctly create the record and roll it back afterwards.

### What actually happened?

The following error occured:

` ActiveRecord::StatementInvalid:
     Mysql2::Error: SAVEPOINT active_record_1 does not exist: ROLLBACK TO SAVEPOINT active_record_1`

### Possible solutions

This error occurs when ActiveRecord tries to rollback to a SAVEPOINT which doesn't exists.
After some checks I've figure out that in https://github.com/test-prof/test-prof/blob/master/lib/test_prof/ext/active_record_3.rb#L8 
we always call `increment_open_transactions `, this leads to always reaching https://github.com/test-prof/test-prof/blob/master/lib/test_prof/ext/active_record_3.rb#L10 **no matter if we are within a transaction or not**. 
Due to the above this method incorrectly creates a `savepoint` when no transactions are currently running which leads to the invalid savepoint issue.
I've monkey patched this locally but just moving line 8 below the `if open_transactions > 0` check and the issue was solved.

If you agree on my proposal I'll make a PR.

intrip:
# What is the purpose of this pull request?

When usig `let_it_be` in Rails3 outside of a transaction the following error appears: `ActiveRecord::StatementInvalid: Mysql2::Error: SAVEPOINT active_record_1 does not exist: ROLLBACK TO SAVEPOINT active_record_1`

The issue is that `increment_open_transactions` is called before `if open_transactions > 0` check which implies that a new transaction is never created.
This PR fixes the issue by moving  `increment_open_transactions` within the if statement: this allows the creation of a transaction when is needed.

Fixes #197

### DIFF
--- a/lib/test_prof/ext/active_record_3.rb
+++ b/lib/test_prof/ext/active_record_3.rb
@@ -5,8 +5,8 @@ module TestProf
   module ActiveRecord3Transactions
     refine ::ActiveRecord::ConnectionAdapters::AbstractAdapter do
       def begin_transaction(joinable: true)
-        increment_open_transactions
         if open_transactions > 0
+          increment_open_transactions
           create_savepoint
         else
           begin_db_transaction


intrip:
@palkan could you please review? I didn't really add any new test here, hope it's allright 🤷‍♂️

PR #205, opened 2021-02-12
Updated call to reporting_enabled?

corroded:
I don't specifically use AnyFixtures so I was wondering why it was suddenly popping up whenever I ran my specs:

image

This PR fixes that outdated call.

After:

image

I followed the deprecation message and updated the call to use the config object instead of the [deprecated] class method

### Checklist
I don't think it's needed since it's an internal change, but let me know if it is :) 


corroded:
P.S.

I've looked into the failing 2.7 build but not sure why. It's a bundler conflict issue?

palkan (approved):
Thanks!

Bundler fails for Rails master (which is 7.0, and some deps have no idea about it 🙂).

PR #209, opened 2021-03-29
Require tsort library

adam12:

With a non-Rails project, minitest, test-prof, and ruby-prof, and invoking tests with `TEST_RUBY_PROF=1 rake test`, I'd receive a `NameError` for `TSort`.

```
/usr/local/bundle/gems/test-prof-1.0.2/lib/test_prof/ruby_prof.rb:243:in `exclude_common_methods': uninitialized constant #<Class:TestProf::RubyProf>::TSort (NameError)
```

Require `tsort` library at top of file using `TSort`.

None.

### Checklist

- I've added tests for this change - no, but open to this one if you feel it's worth it.
- ~I've added a Changelog entry~ NA?
- ~I've updated a documentation~ NA?

### DIFF
--- a/lib/test_prof/ruby_prof.rb
+++ b/lib/test_prof/ruby_prof.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "tsort"
+
 module TestProf
   # RubyProf wrapper.
   #


palkan:
Thanks!

I wonder maybe it makes sense to check for `defined?(TSort)` [here](https://github.com/test-prof/test-prof/pull/209/files#diff-b802c63c2cb57e861605e095e461a4cbf49cc2f85c39b0555dbec141f71a28b2R244) instead?

adam12:
> I wonder maybe it makes sense to check for `defined?(TSort)` [here](https://github.com/test-prof/test-prof/pull/209/files#diff-b802c63c2cb57e861605e095e461a4cbf49cc2f85c39b0555dbec141f71a28b2R244) instead?

Are you thinking you'd like to avoid the overhead of `require`? if not necessary?

palkan:
> avoid the overhead of require if not necessary?

Yeah; but I think, I'm a bit bothered by the _mental_ overhead: why do we require TSort and do not really use it? (That was my first thought: "Wow, I don't remember using TSort in TestProf, I need to refresh my knowledge of the codebase 🙂"

PR #213, opened 2021-05-15
Add support for (process) parallelized tests to before_all (minitest)

peret:
I tried running a test suite using the before_all hook in parallel (using processes) [and it didn't work as intended.](https://github.com/test-prof/test-prof/discussions/211) This PR attempts to fix that. To reiterate my comment from that thread, the issue is that during parallelized execution, an overridden run-method is used to enqueue testing  jobs in a thread-/process-safe manner and this method returns immediately after enqueuing, before any of the tests are run. Therefore, we mustn't roll back the transaction at that point.

This change is based on a monkey-patch for the class `ActiveSupport::Testing::Parallelization::Server`, which wraps a process-safe queue that Rails uses to distribute the test cases. It adds a simple list of testing jobs that are being pushed and popped and after each test case, checks the _next_ job in the queue. If it is of a different class than the class the module was included in, we know that all test cases of that class have finished (or will be by another process) and can roll back the transaction. This works, because this is a FIFO queue and the test cases from each file are pushed one by one, i.e. tests cases from different files are not mixed in the order of the queue.

This change only handles parallelization via processes and more work is needed to also support threads. I might have a stab at that later, but this is my main use-case right now, so I would be glad to see this merged already.

I'm open to any suggestions of how to make this change simpler/more elegant, e.g. how to achieve this without the monkey-patch. Also, I only tested this with my current Rails version, 6.0.3.6, so far.

### Checklist

- [❌] I've added tests for this change
- [✅] I've added a Changelog entry
- [❌] I've updated a documentation

### DIFF
--- a/lib/test_prof/recipes/minitest/before_all.rb
+++ b/lib/test_prof/recipes/minitest/before_all.rb
@@ -6,10 +6,10 @@ Minitest.singleton_class.prepend(Module.new do
   @previous_klass = nil
 
   def run_one_method(klass, method_name)
-    return super unless klass.before_all_executor && klass.parallelized
+    return super unless klass.parallelized
 
     if @previous_klass && @previous_klass != klass
-      klass.before_all_executor.deactivate!
+      @previous_klass.before_all_executor&.deactivate!
     end
     @previous_klass = klass
 
--- a/lib/test_prof/recipes/minitest/before_all.rb
+++ b/lib/test_prof/recipes/minitest/before_all.rb
@@ -100,11 +100,13 @@ module TestProf
         def included(base)
           base.extend ClassMethods
 
-          base.singleton_class.cattr_accessor :parallelized
-          base.parallelize_teardown do
-            last_klass = ::Minitest.previous_klass
-            if last_klass&.respond_to?(:parallelized) && last_klass.parallelized
-              last_klass.before_all_executor&.deactivate!
+          base.cattr_accessor :parallelized
+          if base.respond_to?(:parallelize_teardown)
+            base.parallelize_teardown do
+              last_klass = ::Minitest.previous_klass
+              if last_klass&.respond_to?(:parallelized) && last_klass&.parallelized
+                last_klass.before_all_executor&.deactivate!
+              end
             end
           end
 
--- a/lib/test_prof/recipes/minitest/before_all.rb
+++ b/lib/test_prof/recipes/minitest/before_all.rb
@@ -110,25 +110,27 @@ module TestProf
             end
           end
 
-          base.singleton_class.prepend(Module.new do
-            def parallelize(workers: :number_of_processors, with: :processes)
-              # super.parallelize returns nil when no parallelization is set up
-              if super(workers: workers, with: with).nil?
-                return
-              end
+          if base.respond_to?(:parallelize)
+            base.singleton_class.prepend(Module.new do
+              def parallelize(workers: :number_of_processors, with: :processes)
+                # super.parallelize returns nil when no parallelization is set up
+                if super(workers: workers, with: with).nil?
+                  return
+                end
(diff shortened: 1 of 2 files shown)


peret:
I just realized, there's still a scenarion in which this could fail :-/ I.e., I think it's not guaranteed that the process receives the exact testing job that it sees in `after_teardown`. So there could be a case where the next job is checked, it's the last test from the same class, so the transaction is not rolled back, but another process snatches that job, and the transaction in the current process is never rolled  back. I haven't seen that happen so far, but I guess it's possible.

I'll see if I can come up with a solution for that, but any ideas are definitely appreciated.

peret:
Alright, I changed the implementation so it a) doesn't have the issue of a race condition in detecting the last test job of a test file anymore and b) should be much simpler and (I think) more stable, because it doesn't rely on re-opening and modifying an internal Rails class, that already changed a lot.
Instead, I now intercept calls to `Minitest.run_one_method` to detect when the last test case of a test file has been run, per process.

Could you please let me know if you are willing to merge this change soon, or what you still need from me for this to make it into the next version? :)

lib/test_prof/recipes/minitest/before_all.rb
```diff
@@ -2,6 +2,21 @@
...
 require "test_prof/before_all"
 
+Minitest.singleton_class.prepend(Module.new do
+  @previous_klass = nil
+
+  def run_one_method(klass, method_name)
+    return super unless klass.parallelized
+
+    if @previous_klass && @previous_klass != klass
+      @previous_klass.before_all_executor&.deactivate!
```
palkan, line 12: So, we call `deactivate!` every time a new test class is encountered. But how do we call it for the last test class?

lib/test_prof/recipes/minitest/before_all.rb
```diff
@@ -2,6 +2,21 @@
 
 require "test_prof/before_all"
 
+Minitest.singleton_class.prepend(Module.new do
+  @previous_klass = nil
+
+  def run_one_method(klass, method_name)
+    return super unless klass.parallelized
```
palkan, line 9: We need `klass.respond_to?(:parallelized)` here as well.

lib/test_prof/recipes/minitest/before_all.rb
palkan: If we find a way to perform the last `deactivate!`, we could probably get rid of all the patches and only rely on this logic for both parallelized and non-parallelized scenarios (as far as I understand).

peret: That's a fair point! I've added a `parallelize_teardown` hook to clean up the last transactions in my latest commit.

As far as using this logic for non-parallelized tests as well: That's an interesting idea. I can certainly look into which other patches we can drop. But since `parallelize_teardown` is not called for non-parallelized tests, we would have to keep at least some of the existing patches around I gues.

palkan: Could you please take a look at test failures?

peret: Alright, test cases should be fixed.
While doing that, I also realized that this will probably only work in Rails, not with pure Minitest. I.e. it depends on the tests inheriting from `ActiveSupport::TestCase`. Is that an issue?

palkan: Let's add a check for the presence of the `parallelize` method and we should be good to go (we must do that anyway since we use `super` in the prepended module).

Thus, if no `parallelize` method defined, we do nothing.

peret: Done 👍

peret:
Done ✅

PR #219, opened 2021-07-30
Allow access to `let_it_be` defined variables in after(:context)

cbarton:
Fix accessing `let_it_be` defined variables in after context hooks

[This commit](https://github.com/test-prof/test-prof/commit/250346445396bcd892dae7d8005111a907848498) changed the way we detect the context for fetching the `let_it_be`
defined variables, matching only the `before(:context)`. Those variables
should be accessible by both before/after, so this adds back the generic
`:context` matcher to support all hooks.

### Checklist

### DIFF
--- a/lib/test_prof/recipes/rspec/let_it_be.rb
+++ b/lib/test_prof/recipes/rspec/let_it_be.rb
@@ -107,7 +107,7 @@ module TestProf
         define_method(identifier) do
           # Trying to detect the context
           # Based on https://github.com/rspec/rspec-rails/commit/7cb796db064f58da7790a92e73ab906ef50b1f34
-          if @__inspect_output.include?("before(:context)") || @__inspect_output.include?("before_all")
+          if /(before|after)\(:context\)/.match?(@__inspect_output) || @__inspect_output.include?("before_all")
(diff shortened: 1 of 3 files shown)


lib/test_prof/recipes/rspec/let_it_be.rb
```diff
@@ -107,7 +107,7 @@ def let_it_be(identifier, **options, &block)
         define_method(identifier) do
           # Trying to detect the context
           # Based on https://github.com/rspec/rspec-rails/commit/7cb796db064f58da7790a92e73ab906ef50b1f34
-          if @__inspect_output.include?("before(:context)") || @__inspect_output.include?("before_all")
+          if /\(:context\)/.match?(@__inspect_output) || @__inspect_output.include?("before_all")
```
palkan, line 110: Maybe, we should be more strict here, e.g., `/(before|after)\(:context\)/`?

cbarton: Ya, I can dig that.. squashed in + pushed up.

palkan:
Thanks!

PR #227, opened 2021-12-10
Fix AnyFixture SQL Dump with Newlines

sbader:
Newlines were being stripped from values when stored using `AnyFixture` SQL dumps. This was happening [here](https://github.com/test-prof/test-prof/blob/7b89b548e50a509c528ea02787348f5499f07bd0/lib/test_prof/any_fixture/dump.rb#L66). 

In an effort to maintain the existing functionality and the format of the file, I left that above referenced line in place and replaced all newlines in values with `char(10)` or `chr(10)` depending on the database.

I updated the test to include a newline, which reproduces the issue.

### Checklist


palkan:
Thanks!

PR #229, opened 2022-01-07
JRuby is cat friendly

headius:
All recent versions of JRuby support unicode identifiers.

Does this need tests, changelog, and/or documentation?


palkan:
Awesome 😸 

> Does this need tests, changelog, and/or documentation?

No need, thanks; the existing tests should cover this.

headius:
Thank you!

PR #237, opened 2022-02-15
Respect the printer on the configure block

grillermo-everlyhealth:
I tried customizing the FactoryProf printer and then calling rspec with
`FPROF=1 rspec ` but the custom printer configuration was not being honored, the default printer was being called, this PR fixes that problem.

```ruby
class CustomPrinter
  def self.dump(results, start_time:)
  end
end

TestProf::FactoryProf.configure do |config|
  config.printer = CustomPrinter
end
```

I moved the code being called at exit so it could be tested independently because it is very hard to test code defined at_exit with rspec without a major refactor. 
Then moved the printer definition into that method so if in the mean time a printer is defined on a config block it is actually used. For example if it is defined on a spec_helper or a before block.

The documentation does not need updating, now what is documented is actually working.

### Checklist

- [] I've updated a documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,6 @@
 
 ## master (unrealeased)
 
-## 1.0.8 (2022-02-15
-
 - Fixes the configuration of a printer for factory_prof runs
 
 ## 1.0.7 (2021-08-30)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unrealeased)
 
+## 1.0.8 (2022-02-15
+
+- Fixes the configuration of a printer for factory_prof runs
+
 ## 1.0.7 (2021-08-30)
 
 - Fix access to `let_it_be` variables in `after(:all)` hook. ([@cbarton][])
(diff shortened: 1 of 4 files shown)


grillermo-everlyhealth:
Sorry i did not raise an issue before I thought I understood the problem well enough to propose a fix.

grillermo-everlyhealth:
Fixed the rubocop warnings on the latest commit

lib/test_prof/version.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module TestProf
-  VERSION = "1.0.7"
+  VERSION = "1.0.8"
```
palkan, line 4: Please, do not update the version in the PR. It should be done during release

palkan (approved):
Thanks!
Please, take a look at the version comment.

The CI failures seem irrelevant, so ignore them.

lib/test_prof/version.rb
grillermo-everlyhealth: Oh, of course, I removed the commit that added this change.

PR #236, opened 2022-02-11
restore the lock_thread value after rollback

cou929:
Fixes #232

### What is the purpose of this pull request?

Enables testing of database connections from multiple threads in test cases that do not use before_all.

### What changes did you make? (overview)

Restore the value of `ActiveRecord::Base.connection.pool.lock_thread` after rollback.

### Checklist

- [x] I've added tests for this change
- [x] I've added a Changelog entry
- [ ] I've updated a documentation

PR #249, opened 2022-09-19
don't overwrite saved lock_thread value in nested contexts

ygelfand:
In nested conexts, e.g. :
```
describe 'foo' do
  let_it_be(:foo) { 'bar' }
  describe 'bar' do
    let_it_be(:baz) { 'bax' }
    it do
      expect(true).to be_truthy
    end
  end
end
```

the saved `_orig_lock_thread` value introduced in https://github.com/test-prof/test-prof/pull/236 gets overriden, and is only restored back to `true`, this avoids re-saving thevalue if it was already saved

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+- Fixed restoring lock_thread value in nested contexts ([@ygelfand][])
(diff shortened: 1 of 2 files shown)


palkan:
Great! Thanks

PR #252, opened 2022-10-05
Disable GC frames with TEST_STACK_PROF_IGNORE_GC=1

cbliard:

Add option `ignore_gc: true` when calling `StackProf.run` if `TEST_STACK_PROF_IGNORE_GC` is set.

As per StackProf README:
> By default, samples taken during garbage collection will show as garbage collection frames including both mark and sweep phases. For longer traces, these can leave gaps in a flamegraph that are hard to follow. They can be disabled by setting the `ignore_gc` option to true. Garbage collection time will still be present in the profile but not explicitly marked with its own frame.

When `TEST_STACK_PROF_IGNORE_GC` env var is set, the `ignore_gc: true` option is added.

No

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+- Disable garbage collection frames when `TEST_STACK_PROF_IGNORE_GC` env variable is set ([@cbliard][])
+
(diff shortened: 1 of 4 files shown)


palkan:
Thanks!

Perfect PR 👍

Issue #234, opened 2022-02-07
Global before_all tags support


palkan:
## Use Cases

Such a filtering can be obviously used for debugging. Another use case is adding an exception for the cases when you shouldn't build any objects inside the transaction. Let's presume we have some groups tagged as which means that before each example we would reset all the database sequences so that newly created objects started their ids from 1. In this case we probably shouldn't build objects in (neither define them using ) because ids of these long-living objects can get into a conflict with those created inside an example-specific transaction. We could either enforce this, or just add the warning message like:

```ruby
TestProf::BeforeAll.configure do |config|
  config.before(:begin, reset_sequences: true) do
    warn <<~MESSAGE
      Do NOT create objects outside of transaction
      because all db sequences will be reset to 1
      in every single example, so that IDs of new objects
      can get into conflict with the long-living ones.
    MESSAGE
  end
end
```

## Reference

~~Please look at the [examples of RSpec filtering](https://relishapp.com/rspec/rspec-core/v/3-10/docs/hooks/filters) and check how they are implemented (possibly starting from[ here](https://github.com/rspec/rspec-core/blob/main/lib/rspec/core/metadata_filter.rb)). 
Ideally, the gem doesn't need to implement the filtering by itself, but somehow reuse the existing feature of RSpec. This would open doors for wealthier features (like filtering with lambdas etc.) and for better support in the future.~~

---

Proposed by @nepalez.

palkan:
Closed by #255

PR #255, opened 2022-11-25
Global before_all tags support

maxshend:

To add support for global before_all tags mentioned in #234. It will allow to register callbacks for specific RSpec Example Groups using tags.
```ruby
TestProf::BeforeAll.configure do |config|
  config.before(:begin, reset_sequences: proc(&:present?)) do
    warn <<~MESSAGE
      Do NOT create objects outside of transaction
      because all db sequences will be reset to 1
      in every single example, so that IDs of new objects
      can get into conflict with the long-living ones.
    MESSAGE
  end
end
```

### Checklist


CHANGELOG.md
```diff
@@ -2,6 +2,20 @@
 
 ## master (unreleased)
 
+- Add support for global `before_all` tags
+```ruby
+TestProf::BeforeAll.configure do |config|
+  config.before(:begin, reset_sequences: proc(&:present?)) do
```
palkan, line 8: Can we use RSpec API for specifying tags? I mean:

```ruby
config.before(:begin, reset_sequences: true) do
end

# or
config.before(:begin, type: :system) do
end
```

?

We need to support both presence (`true`) and equality.

UPD: I see that using RSpec utils was in the proposal; I don't think we need to be 100% compatible with RSpec, the feature should cover common cases (and common sense, which is not passing procs as Hash keys). Let's make it simpler.

palkan (approved):
Perfect!

Could you please add a change log entry?

CHANGELOG.md
```diff
@@ -33,6 +33,20 @@ end
 
 You can also use `TestProf::FactoryDefault.disable!(&block)`.
 
+- Add support for global `before_all` tags ([@maxshend][])
```
maxshend, line 36: > Could you please add a change log entry?

@palkan I think I've already added a log entry. Do I need to change anything else ? Thank you

palkan: I'm sorry, confused it with documentation update.

All good 👍

PR #257, opened 2022-11-28
Add support for multiple databases to BeforeAll ActiveRecord adapter

rutgerw:

Add support for ActiveRecord multiple databases

- The `BeforeAll` ActiveRecord adapter will now call `begin_transaction` on all descendants of ActiveRecord::Base.
- Added specs for the adapter class

### Checklist

- [X] I've added tests for this change
- [X] I've updated a documentation

### DIFF
--- a/lib/test_prof/before_all/adapters/active_record.rb
+++ b/lib/test_prof/before_all/adapters/active_record.rb
@@ -6,17 +6,27 @@ module TestProf
       # ActiveRecord adapter for `before_all`
       module ActiveRecord
         class << self
+          def all_connection_classes
+            @all_connection_classes ||= [::ActiveRecord::Base] + ::ActiveRecord::Base.descendants.select do |descendant|
+              descendant&.connection_class?
+            end.compact
+          end
+
           def begin_transaction
-            ::ActiveRecord::Base.connection.begin_transaction(joinable: false)
+            all_connection_classes.each do |connection_class|
+              connection_class.connection.begin_transaction(joinable: false)
+            end
           end
 
           def rollback_transaction
-            if ::ActiveRecord::Base.connection.open_transactions.zero?
-              warn "!!! before_all transaction has been already rollbacked and " \
-                   "could work incorrectly"
-              return
+            all_connection_classes.each do |connection_class|
+              if connection_class.connection.open_transactions.zero?
+                warn "!!! before_all transaction has been already rollbacked and " \
+                      "could work incorrectly"
+                next
+              end
+              connection_class.connection.rollback_transaction
             end
-            ::ActiveRecord::Base.connection.rollback_transaction
           end
 
           def setup_fixtures(test_object)
--- a/lib/test_prof/before_all/adapters/active_record.rb
+++ b/lib/test_prof/before_all/adapters/active_record.rb
@@ -6,26 +6,24 @@ module TestProf
       # ActiveRecord adapter for `before_all`
       module ActiveRecord
         class << self
-          def all_connection_classes
-            @all_connection_classes ||= [::ActiveRecord::Base] + ::ActiveRecord::Base.descendants.select do |descendant|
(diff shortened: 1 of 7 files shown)


lib/test_prof/before_all/adapters/active_record.rb
```diff
@@ -6,17 +6,27 @@ module Adapters
       # ActiveRecord adapter for `before_all`
       module ActiveRecord
         class << self
+          def all_connection_classes
+            @all_connection_classes ||= [::ActiveRecord::Base] + ::ActiveRecord::Base.descendants.select do |descendant|
```
palkan, line 10: Maybe, we can use the AR configuration here? Similarly to what Rails does for fixtures: https://github.com/rails/rails/blob/fa09e3c6787d8dd5f3eff30e07deb67b249af6b6/activerecord/lib/active_record/test_fixtures.rb#L110-L114

spec/test_prof/before_all/adapters/active_record_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+module TestProf
+  module BeforeAll
+    def self.configure
+    end
+  end
+end
+
+require "test_prof/before_all/adapters/active_record"
+
+describe "TestProf::BeforeAll::Adapters::ActiveRecord" do
```
palkan, line 12: It would be better (and I think easier and more useful in the long term) to write an integration test with real Active Record multi-db setup. What do you think?

palkan (changes requested):
Thanks!

This is a great addition to `before_all`.

Left some comments and suggestions.

spec/test_prof/before_all/adapters/active_record_spec.rb
rutgerw: I tried it first but I ran into issues... but now I have a working setup, I can give it another shot.

rutgerw: I struggled with the AR setup. I am now using an environment variable `MULTI_DB`, if true, an additional table `comments` is added in a separate sqlite database. 

If it is an OK solution, should we then run all variations in `rspec.yml` for single and multiple db setup? WDYT?

Alternative setup:
- Duplicate the AR models setup: 1 for Rails 5 (single db)and 1 for Rails 6 and up (multiple db)

palkan: We don't need all the variations; let's run tests with `MULTI_DB=true` for Rails 7/master and `MULTI_DB=false` otherwise.

lib/test_prof/before_all/adapters/active_record.rb
```diff
@@ -6,17 +6,29 @@ module Adapters
       # ActiveRecord adapter for `before_all`
       module ActiveRecord
         class << self
+          def all_connections
+            @all_connections ||= if ::ActiveRecord::Base.respond_to? :connects_to
```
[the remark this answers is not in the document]
rutgerw: Yeah makes sense, I added it.

spec/test_prof/before_all/adapters/active_record_spec.rb
rutgerw: Sure, changed it

rutgerw:
Noticed that the failures are related to failure in package fetching `E: Unable to locate package postgresql-client-12`

palkan:
Hm, tests are failing:

- [`wrong number of arguments (given 1, expected 0)` in connection_pool_list](https://github.com/test-prof/test-prof/actions/runs/3839563629/jobs/6537499103)
- [`Could not find table 'users'`](https://github.com/test-prof/test-prof/actions/runs/3839563629/jobs/6537498616) (probably, a different database is used?)


Could you please take a look?

rutgerw:
Yes I will

PR #260, opened 2023-01-05
For multi db, use ApplicationRecord to create connection

rutgerw:
After merge of Multi DB, some tests were failing. 

This change fixed it locally. I ran different combinations of rspec locally.


palkan:
Thanks!

PR #262, opened 2023-02-08
Add FactoryBot.get_factory_default

john-h-k:
Adds a `FactoryBot.get_factory_default` method that allows users to retrieve the default for a given setup. Useful in cases where the factory default is desired but the record isn't directly created through factory bot (such as an after build hook)

### DIFF
--- a/lib/test_prof/factory_default.rb
+++ b/lib/test_prof/factory_default.rb
@@ -168,6 +168,10 @@ module TestProf
         )
       end
 
+      def get_factory_default(name, *args)
+        FactoryDefault.get(name, traits = args, overrides = args.extract_options!, write_stats = false)
+      end
+
       def skip_factory_default(&block)
         FactoryDefault.disable!(&block)
       end
@@ -236,7 +240,7 @@ module TestProf
         obj
       end
 
-      def get(name, traits = nil, overrides = nil)
+      def get(name, traits = nil, overrides = nil, write_stats = true)
         return unless enabled?
 
         record = store[name]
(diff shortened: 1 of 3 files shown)


lib/test_prof/factory_default.rb
```diff
@@ -168,6 +168,10 @@ def set_factory_default(name, obj, preserve_traits: FactoryDefault.config.preser
         )
       end
 
+      def get_factory_default(name, *args)
```
palkan, line 171: Let's call it `#fetch_factory_default`

lib/test_prof/factory_default.rb
```diff
@@ -168,6 +168,10 @@ def set_factory_default(name, obj, preserve_traits: FactoryDefault.config.preser
...
+        FactoryDefault.get(name, args, args.extract_options!, false)
```
palkan, line 172: Why can't we use `*args, **kwargs` instead of `#extract_options!`? No need to support legacy Hash-to-kwarg conversion, it's a new API method.

lib/test_prof/factory_default.rb
```diff
@@ -236,7 +240,7 @@ def register(name, obj, **options)
         obj
       end
 
-      def get(name, traits = nil, overrides = nil)
+      def get(name, traits = nil, overrides = nil, write_stats = true)
```
palkan, line 243: Let's rename to `skip_stats` and turn this option into a keyword argument.

palkan (changes requested):
Thanks for the PR!

This is useful feature; let's just polish the API a bit (see the comments).

PR #267, opened 2023-06-16
feat: easily ignore pools which cannot cannot connect

bf4:

For an application with multiple databases, allow the databases not-under-test to not be running (not connectable) and for before_all to still work.

Our primary database is postgres, but we also have a sql server database and some others.  Prior to this PR, before_all fails if the SQL Server database isn't running, even if it isn't under test.

This PR makes it possible to opt out of using before_all for databases with known pool connection errors.

I didn't write any tests. I'm happy to. This PR is kind of proof of concept, or PR as issue.

I tried to make it easy to extend the default behavior without a monkey patch.

### Checklist

### DIFF
--- a/lib/test_prof/before_all/adapters/active_record.rb
+++ b/lib/test_prof/before_all/adapters/active_record.rb
@@ -8,10 +8,25 @@ module TestProf
         class << self
           def all_connections
             @all_connections ||= if ::ActiveRecord::Base.respond_to? :connects_to
-              ::ActiveRecord::Base.connection_handler.connection_pool_list.map(&:connection)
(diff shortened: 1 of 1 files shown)


bf4:
Test failures seem upstream of this change?

palkan:
> Test failures seem upstream of this change?

Yeah, just fixed in master.


Merging. Thanks!

bf4:
Oh, awesome!  I'll remove our patch and use this when released

palkan:
Released in 1.2.2.

bf4:
we're now using this 👯 

```ruby
require "test_prof/before_all/adapters/active_record"
TestProf::BeforeAll::Adapters::ActiveRecord
  .pool_connection_errors << TinyTds::Error
```

PR #277, opened 2023-10-27
Introduce MemoryProf for detecting memory spikes

Vankiru:

Introducing a memory profiler `MemoryProf` that can help to detect test examples and groups that cause memory spikes. Memory profiling supports two metrics: RSS and allocations.
 
Examples:
```sh
TEST_MEM_PROF='rss' rspec ...
TEST_MEM_PROF='alloc' rspec ...
```

By default MemoryProf shows the top 5 examples and groups (for RSpec) but you can set how many items to display with:
```sh
 TEST_MEM_PROF='rss' TEST_MEM_PROF_COUNT=10 rspec ...
```

The examples block shows the amount of memory used by each example, and the groups block displays the memory allocated by other code defined in the groups. For example, RSpec groups may include heavy `before(:all)` (or `before_all`) setup blocks, so it is helpful to see which groups use the most amount of memory outside of their examples.

`MemoryProf` has the following structure:

1. Configuration defines MemoryProf configs and adds a few supporting methods ([memory_prof.rb](https://github.com/Vankiru/test-prof/blob/feature/memory-prof/lib/test_prof/memory_prof.rb)).
2. Trackers are responsible for tracking memory usage and determining the top n examples and groups ([memory_prof/tracker.rb](https://github.com/Vankiru/test-prof/blob/feature/memory-prof/lib/test_prof/memory_prof/tracker.rb)).
4. Printer assembles and prints the results of a tracker ([memory_prof/printer.rb](https://github.com/Vankiru/test-prof/blob/feature/memory-prof/lib/test_prof/memory_prof/printer.rb)).
5. RSpec integration ([memory_prof/rspec.rb](https://github.com/Vankiru/test-prof/blob/feature/memory-prof/lib/test_prof/memory_prof/rspec.rb)).
6. Minitest integration ([memory_prof/minitest.rb](https://github.com/Vankiru/test-prof/blob/feature/memory-prof/lib/test_prof/memory_prof/minitest.rb)).

There are two types of trackers: `AllocTracker` and `RssTracker`. A tracker consists of four main parts:
* `list` - a linked list that is used to track memory for individual groups/examples, and it is an instance of `LinkedList `(for more info see ([memory_prof/tracker/linked_list.rb](https://github.com/Vankiru/test-prof/blob/feature/memory-prof/lib/test_prof/memory_prof/tracker/linked_list.rb)).
* `examples` – the top n examples, an instance of `Utils::SizedOrderedSet`.
* `groups` – the top n groups, an instance of `Utils::SizedOrderedSet`.
* `track` - a method that fetches the amount of memory in use at a certain point.

Since RSS depends on the host OS, a separate tool (`RssTool`) is introduced to handle how to fetch RSS for a specific operating system ([rss_tool.rb](https://github.com/Vankiru/test-prof/blob/feature/memory-prof/lib/test_prof/memory_prof/tracker/rss_tool.rb)). There are 3 RSS tools currently available:

* `ProcFS` – reads `/proc/$pid/statm` file (Linux).
*  `PS` – uses `ps` (macOS, Solaris, BSD).
* `GetProcess` – uses `Get-Process`, requires PowerShell to be installed (Windows).

In a similar way to trackers, there are two types of printers: `AllocPrinter` and `RssPrinter`. Their implementations differ only in a few methods that are responsible for formatting memory values.

### Notes on JRuby

1. JRuby `GC.stat` returns a very different set of values that doesn't include anything like `total_allocated_objects ` in MRI (which is fine cause the implementations are different). I googled a bit about how to get allocations count on JRuby but did not find anything useful. So because of that the allocation mode is disabled on JRuby for now.
2. It seems that the JRuby build fails for some reasons, even in master. I tried a few JRuby versions but it didn't help. I run the specs locally and they all pass. Here is the link to one of the builds: https://github.com/Vankiru/test-prof/actions/runs/6651407791/job/18073378978

### Checklist
 
### DIFF
--- a/docs/profilers/memory_prof.md
+++ b/docs/profilers/memory_prof.md
@@ -1,0 +1,85 @@
+# MemoryProf
+
+MemoryProf tracks memory usage during your test suite run, and can help to detect test examples and groups that cause memory spikes. Memory profiling supports two metrics: RSS and allocations.
+
+Example output:
+
+```sh
+[TEST PROF INFO] MemoryProf results
+
+Final RSS: 673KB
+
+Top 5 groups (by RSS):
+
+AnswersController (./spec/controllers/answers_controller_spec.rb:3) – +80KB (13.50%)
+QuestionsController (./spec/controllers/questions_controller_spec.rb:3) – +32KB  (9.08%)
+CommentsController (./spec/controllers/comments_controller_spec.rb:3) – +16KB (3.27%)
+
+Top 5 examples (by RSS):
+
+destroys question (./spec/controllers/questions_controller_spec.rb:38) – +144KB (24.38%)
+change comments count (./spec/controllers/comments_controller_spec.rb:7) – +120KB (20.00%)
+change Votes count (./spec/shared_examples/controllers/voted_examples.rb:23) – +90KB (16.36%)
+change Votes count (./spec/shared_examples/controllers/voted_examples.rb:23) – +64KB (12.86%)
+fails (./spec/shared_examples/controllers/invalid_examples.rb:3) – +32KB (5.00%)
+```
+
+The examples block shows the amount of memory used by each example, and the groups block displays the memory allocated by other code defined in the groups. For example, RSpec groups may include heavy `before(:all)` (or `before_all`) setup blocks, so it is helpful to see which groups use the most amount of memory outside of their examples.
+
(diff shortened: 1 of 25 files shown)


palkan:
Thanks! Fantastic work 👍

PR #283, opened 2023-12-22
Add Minitest support to TagProf

lioneldebauge:

Following this [discussion](https://github.com/test-prof/test-prof/discussions/272#discussion-5537313) this PR aims to add Minitest support for TagProf.

Created a `TagProfReporter` class which reproduces `TestProf::TagProf::RSpecListener` behavior. The main differences are:
- because tags do not exist in Minitest we use subdirectories of main test directory to group tests statistics
- when a main test directory cannot be determined it groups by test and display a significant warning message

You can read the comments for my different questions :) 

### Checklist

### DIFF
--- a/docs/profilers/tag_prof.md
+++ b/docs/profilers/tag_prof.md
@@ -35,15 +35,17 @@ TagProf can be used with both RSpec and Minitest (limited support, see  below).
 To activate TagProf use `TAG_PROF` environment variable:
 
 With Rspec:
+
 ```sh
 # Group by type
 TAG_PROF=type rspec
 ```
 
 With Minitest:
+
 ```sh
 # using pure ruby
-TAG_PROF=type ruby 
+TAG_PROF=type ruby
 
 # using Rails built-in task
 TAG_PROF=type bin/rails test
@@ -56,14 +58,15 @@ Minitest does not support the usage of tags by default. TagProf therefore groups
 When no root test directory can be found the test statistics will not be grouped with other tests. They will be displayed per test with a significant warning message in the report.
 
 Example:
+
 ```sh
 [TEST PROF INFO] TagProf report for type
 
-       type          time   sql.active_record  total  %total   %time           avg
+type          time   sql.active_record  total  %total   %time           avg
 
-    subdirectory_not_found: subdirectory_not_found_fixture.rb     00:00.101       1  100.00  100.00     00:00.101
- controller     00:02.855           00:00.921     42   33.87   32.48     00:00.067
-      model     00:01.127           00:00.446     40   32.26   12.82     00:00.028
+subdirectory_not_found: subdirectory_not_found_fixture.rb     00:00.101       1  100.00  100.00     00:00.101
+controller     00:02.855           00:00.921     42   33.87   32.48     00:00.067
+model     00:01.127           00:00.446     40   32.26   12.82     00:00.028
 ```
 
 ## Profiling events
--- a/docs/profilers/tag_prof.md
+++ b/docs/profilers/tag_prof.md
@@ -51,6 +51,8 @@ TAG_PROF=type ruby
 TAG_PROF=type bin/rails test
 ```
 
+NB: if another value than "type" is used for TAG_PROF environment variable it will be ignored silently in both Minitest and RSpec.
+
 ### Usage specificity with Minitest
 
 Minitest does not support the usage of tags by default. TagProf therefore groups statistics by direct subdirectories of the root test directory. It assumes root test directory is named either `spec` or `test`.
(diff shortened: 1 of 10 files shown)


docs/profilers/tag_prof.md
```diff
@@ -30,15 +30,42 @@ That's how a report looks like:
 
 ## Instructions
 
-TagProf can only be used with RSpec.
+TagProf can be used with both RSpec and Minitest.
```
lioneldebauge, line 33: I'm not sure TagProf is a very relevant name for Minitest. I was wondering if I should not create another profiler called `DirectoryProf` and refer to it in TagProf documentation as a Minitest equivalent. What do you think ?

lib/test_prof/tag_prof/minitest.rb
```diff
@@ -0,0 +1,72 @@
...
+
+      def report
+        printer = (ENV["TAG_PROF_FORMAT"] == "html") ? ::TestProf::TagProf::Printers::HTML : ::TestProf::TagProf::Printers::Simple
+        printer.dump(results)
+      end
+
+      private
+
+      def main_folder_path(result)
+        return "subdirectory_not_found: #{result.source_location.first}" if absolute_path_from(result).nil?
```
lioneldebauge, line 40: I'm not sure if this is the right behavior. Maybe we want to raise if the root test directory cannot be found ?

docs/profilers/tag_prof.md
palkan: Let's use the `TagProf` name for now. We'll probably add support for other tags beyond "type" in the future (maybe, via integration with some third-party plugins, e.g., `minitest-metadat` or similar).

docs/profilers/tag_prof.md
```diff
@@ -30,15 +30,42 @@ That's how a report looks like:
...
+
+# using Rails built-in task
+TAG_PROF=type bin/rails test
+```
+
+### Usage specificity with Minitest
+
+Minitest does not support the usage of tags by default. TagProf therefore groups statistics by direct subdirectories of the root test directory. It assumes root test directory is named either `spec` or `test`.
+
+When no root test directory can be found the test statistics will not be grouped with other tests. They will be displayed per test with a significant warning message in the report.
```
palkan, line 56: I think, we should use the `__unknown__` value for unrecognized tests (that's how we deal with RSpec specs without defined tags).

palkan: Also, I think, we should make it clear in the docs that if some other value beyound "type" is provided, it's ignored silently.

lib/test_prof/tag_prof/minitest.rb
palkan: Answered above: let's return `:__unknown__`

docs/profilers/tag_prof.md
palkan, line 33:
```suggestion
TagProf can be used with both RSpec and Minitest (limited support, see  below).
```

palkan (changes requested):
Thanks! Looks good.

Left a comment regarding unrecognized paths.

And we also have some CI tests problems—could you please take a look?

gemfiles/railsmaster.gemfile
```diff
@@ -1,6 +1,6 @@
 source "https://rubygems.org"
 
-gem "rails", github: "rails/rails"
+gem "rails", "~> 7.0"
```
lioneldebauge, line 3: This allows to fix [bugs](https://github.com/lioneldebauge/test-prof/actions/runs/7865021215/job/21457479005) related to a constant not being loaded in Rails. I fixed it by using only the last stable branch of version 7 but maybe there is a good reason why we don't do that already ?

Bugs was occuring both on my branch and on master

lioneldebauge:
@palkan thanks for the review. I have triggered CI in my fork with my GA free plan and everything should be green and suggestions/comments should be addressed now. Let me know what you think!

gemfiles/railsmaster.gemfile
palkan: Let's try adding `require "active_support/inflector"` somewhere here (before `factory_bot`): https://github.com/test-prof/test-prof/blob/02d8f355c158fb021e58ff1327d624a8299762b6/spec/support/ar_models.rb#L9

lioneldebauge: @palkan done. CI is all green on my side 👍

palkan (approved):
Amazing work 👍 Thank you!

Will craft a new release soon

mjankowski:
Looks like a rubygems release was pushed, but no changelog/tag yet?

PR #287, opened 2024-04-15
refactor: avoid using activesupport's present? method

### LINKED ISSUE #284: Undefined method #present? when using test-prof w/o activesupport
### Tell us about your environment

**Ruby Version:** 3.2

**Framework Version (RSpec, Minitest, FactoryGirl, Rails, whatever):** RSpec, No Rails

**TestProf Version:** 1.3.2 ()

### What did you do?

I upgraded to test-prof v1.3.2 and ran my tests locally.

### What did you expect to happen?

No test errors

### What actually happened?

I got failing tests because of:

```
Failure/Error: return true unless filters.present? && TestProf.rspec?
     
     NoMethodError:
       undefined method `present?' for {}:Hash
     
               return true unless filters.present? && TestProf.rspec?
                                         ^^^^^^^^^
     # ..../test-prof-1.3.2/lib/test_prof/before_all.rb:69:in `filters_apply?'
```

I suspect this is because I don't have activesupport in my project.

I'm happy to provide a fix to not depend on `.present?` for that particular case, let me know if that's ok.

ngarbezza:

Fixes #284, for cases when test-prof is not used in the context of Rails (activesupport)

Avoid using `present?` and instead solve this usage with Ruby std-lib code. I think `is_a?(Hash)` is kind of equivalent for this purpose.

### Checklist

- [ ] I've added tests for this change: I didn't know how/where to test this. I'm open to suggestions!
- [x] I've updated a documentation: no need for documentation as it's an internal change

-->

### DIFF
--- a/lib/test_prof/before_all.rb
+++ b/lib/test_prof/before_all.rb
@@ -66,7 +66,7 @@ module TestProf
       private
 
       def filters_apply?(metadata)
-        return true unless filters.present? && TestProf.rspec?
+        return true unless filters.is_a?(Hash) && TestProf.rspec?
 
         ::RSpec::Core::MetadataFilter.apply?(
           :all?,


palkan:
Awesome! Thanks!

Issue #291, opened 2024-05-04
Warn on let_it_be duplications


palkan:
## Context

Although we suggest using `let_it_be` instead of `let!`, there is one important difference: you can override `let!` definition with the same or nested context, so only the latter one is called; `let_it_be` records could be overridden, but still created. For example:

```ruby
context "A" do
  let!(:user) { create(:user, name: "a") } 
  let_it_be(:post) { create(:post, title: "A") }

  specify { expect(User.all.pluck(:name)).to eq ["a"] }
  specify { expect(Post.all.pluck(:title)).to eq ["A"] }

  context "B" do
    let!(:user) { create(:user, name: "b") }
    let_it_be(:post) { create(:post, title: "B") }

    specify { expect(User.all.pluck(:name)).to eq ["b"] }
    specify { expect(Post.all.pluck(:title)).to eq ["B"] } # fails, because there are two posts
  end
end
```

The situation becomes even more trickier when you have multiple let_it_be records and dependencies between them.

## TODO

Add a configuration option to warn or raise an exception whenever a let_it_be definition is redefined:

```ruby
TestProf::LetItBe.configure do |config|
  config.report_duplicates = :warn # Kernel.warn
  config.report_duplicates = :raise # Kernel.raise
end
```

PR #293, opened 2024-05-17
Add report duplicates config option for let it be

lHydra:

Add `report_duplicates` config option to avoid surprises when overriding let_it_be in a nested context

I added a stub method, to track overrides (`define_method("let_it_be_#{identifier}") {}`).
But somehow it looks doubtful, maybe I should use another approach?

Also I decided to use the method that comes with rspec (`RSpec.warn_with`) instead of `Kernel.warn`.

Also, I'm not sure if the default option `report_duplicates = :warn` is necessary?

### Checklist

Closes https://github.com/test-prof/test-prof/issues/291

### DIFF
--- a/lib/test_prof/recipes/rspec/let_it_be.rb
+++ b/lib/test_prof/recipes/rspec/let_it_be.rb
@@ -7,6 +7,8 @@ module TestProf
   # Just like `let`, but persist the result for the whole group.
   # NOTE: Experimental and magical, for more control use `before_all`.
   module LetItBe
+    class DuplicationError < StandardError; end
+
     Modifier = Struct.new(:scope, :block) do
       def call(record, config)
         block.call(record, config)
@@ -30,6 +32,19 @@ module TestProf
         raise ArgumentError, "Modifier #{key} is already defined for let_it_be" if LetItBe.modifiers.key?(key)
 
         LetItBe.modifiers[key] = Modifier.new(on, block)
+      end
+
+      def report_duplicates=(value)
+        value = value.to_sym
+        unless %i[warn raise].include?(value)
+          raise ArgumentError, "#{value} is not acceptable, acceptable values are :warn or :raise"
+        end
+
+        @report_duplicates = value
+      end
+
+      def report_duplicates
+        @report_duplicates ||= false
       end
 
       def default_modifiers
@@ -117,6 +132,8 @@ module TestProf
         instance_variable_get(:"#{PREFIX}#{identifier}")
       end
 
+      report_duplicates(identifier) if LetItBe.config.report_duplicates
+
       LetItBe.module_for(self).module_eval do
         define_method(identifier) do
           # Trying to detect the context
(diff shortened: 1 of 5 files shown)


lib/test_prof/recipes/rspec/let_it_be.rb
```diff
@@ -32,6 +34,19 @@ def register_modifier(key, on: :let, &block)
...
+        value = value.to_sym
+        unless %i[warn raise].include?(value)
+          raise ArgumentError, "#{value} is not acceptable, acceptable values are :warn or :raise"
+        end
+
+        @report_duplicates = value
+      end
+
+      def report_duplicates
+        @report_duplicates ||= :warn
```
palkan, line 47: I think, the default should be false meaning that we don't need any warnings (current behaviour).

lib/test_prof/recipes/rspec/let_it_be.rb
```diff
@@ -130,6 +155,9 @@ def let_it_be(identifier, **options, &block)
             super()
           end
         end
+
+        # Used to to track method overrides
+        define_method("let_it_be_#{identifier}") {}
```
palkan, line 160: I don't like this idea. Let's try to use the information we already have to detect if there is already a let_it_be-defined method.

For example, we can check that the `instance_method_defined?(identifier)` and then check the location of the definition, and if it's the current file—it's defined by `let_it_be`. Something like that.

palkan (changes requested):

lib/test_prof/recipes/rspec/let_it_be.rb
lHydra: Ah, I didn't realize that we can check the source file where the method is defined

Done

palkan (approved):
Looks good 👍 Thanks!

Please, rebase, and we can merge it.

lHydra:
@palkan some weird test fails in CI... Can you help me, please?

lib/test_prof/recipes/rspec/let_it_be.rb
```diff
@@ -32,6 +34,19 @@ def register_modifier(key, on: :let, &block)
...
+        value = value.to_sym
+        unless %i[warn raise].include?(value)
+          raise ArgumentError, "#{value} is not acceptable, acceptable values are :warn or :raise"
+        end
+
+        @report_duplicates = value
+      end
+
+      def report_duplicates
+        @report_duplicates ||= false
```
palkan, line 47: This memoization is useless; we can skip setting to false and use attr_reader instead

lib/test_prof/recipes/rspec/let_it_be.rb
```diff
@@ -32,6 +34,19 @@ def register_modifier(key, on: :let, &block)
         LetItBe.modifiers[key] = Modifier.new(on, block)
       end
 
+      def report_duplicates=(value)
+        value = value.to_sym
```
palkan, line 38: What about `report_duplicates = false`? (For example, to temporary disable reporting).

I think, we can remove the value validation altogether and be less defensive here

palkan (commented):
Left a few more comments.

CI failures are unrelated. They caused by the recent Timecop release: https://github.com/travisjeffery/timecop/pull/419

lib/test_prof/recipes/rspec/let_it_be.rb
lHydra: remove excess memoization

lib/test_prof/recipes/rspec/let_it_be.rb
lHydra: > What about report_duplicates = false? (For example, to temporary disable reporting).

We could omit this value, which would be equiv

lHydra: removed the validation

palkan (approved):
Perfect! Thank you!

Issue #290, opened 2024-05-04
Add threshold configuration to FactoryProf


palkan:
## Context

When running FactoryProf (`FPROF=1`), the output may contain a lot of lines for factories that has been used a few times, thus, making it harder to quickly identify overused factories.

We can safely exclude rarely used factories from the output without sacrificing the value of the profile.

## TODO

Let's add a configuration parameter to FactoryProf, `threshold`, which would be used in the profile printer to ignore factories which total number of calls is less than the provided threshold.

It should be possible to set the threshold value via Ruby or env vars (`FPROF_THRESHOLD=N`).

PR #294, opened 2024-05-18
add threshold config parameter for FactoryProf

lHydra:

Add threshold config parameter for `FactoryProf` to ignore factories which total number of calls is less than the provided threshold

I'm not sure if support for this option is needed for flamegraph mode? As of now, I have only added support for simple mode

### Checklist

Closes https://github.com/test-prof/test-prof/issues/290

### DIFF
--- a/docs/profilers/factory_prof.md
+++ b/docs/profilers/factory_prof.md
@@ -73,7 +73,7 @@ Example output:
 [TEST PROF INFO] Profile results to JSON: tmp/test_prof/test-prof.result.json
 ```
 
-### Threshold config
+### Reducing the output
 
 When running FactoryProf, the output may contain a lot of lines for factories that has been used a few times.
 To avoid this and focus on the most important statistics you can specify a threshold value. Then you will be shown the factories whose total number exceeds the threshold.
--- a/docs/profilers/factory_prof.md
+++ b/docs/profilers/factory_prof.md
@@ -119,6 +119,28 @@ The wider column the more often this stack appears.
 
 The `root` cell shows the total number of `create` calls.
 
+## Threshold config
+
+When running FactoryProf, the output may contain a lot of lines for factories that has been used a few times.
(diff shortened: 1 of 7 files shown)


spec/integrations/factory_prof_spec.rb
```diff
@@ -13,6 +13,17 @@
       expect(output).to match(/\s+16\s+8\s+(\d+\.\d{4}s\s+){3}user\n\s+10\s+6\s+\s+(\d+\.\d{4}s\s+){3}post/)
     end
 
+    specify "simple printer with threshold param", :aggregate_failures do
+      output = run_rspec("factory_prof", env: {"FPROF" => "1", "FPROF_THRESHOLD" => "11"})
+
+      expect(output).to include("FactoryProf enabled (simple mode)")
+
+      expect(output).to include("Factories usage")
+      expect(output).to match(/Total: 16\n\s+Total top-level: 8\n\s+Total time: \d{2}+:\d{2}\.\d{3} \(out of \d{2}+:\d{2}\.\d{3}\)\n\s+Total uniq factories: 1/)
```
palkan, line 22: We should only hide less used factories from the output but totals still must show the actual data, i.e., here we should see "Total uniq factories: 2".

Thus, we need to update the exclusion logic (no `select!`, just skip when printing) and update the test to check that there is not match for skipped factories.

docs/profilers/factory_prof.md
```diff
@@ -73,6 +73,28 @@ Example output:
 [TEST PROF INFO] Profile results to JSON: tmp/test_prof/test-prof.result.json
 ```
 
+### Threshold config
```
palkan, line 76:
```suggestion
### Reducing the output
```

palkan (changes requested):
Thanks!

Left a comment regarding the current logic. Let's fix it.

spec/integrations/factory_prof_spec.rb
lHydra: done

palkan (approved):
Perfrect 👍 Thanks!

Issue #289, opened 2024-05-04
Add Rails/ActiveSupport hooks support for Vernier profiles


palkan:
## Context

Vernier allows adding event markers from Active Support Notifications.

This functionality must be explicitly enabled via the `hooks: [:rails]` option for the Collector.

This will add more insights to the resulting report. For example, we can see which initializers are being executed during the application boot and how much time did they take:

<img width="1210" alt="image" src="https://github.com/test-prof/test-prof/assets/1516722/87c37f34-2b4b-4ce2-a0e4-de375bddd54b">

## TODO

Add new configuration parameter to the Vernier wrapper (`test_prof/vernier.rb`) to add `hooks: [:rails]` when initializing the collector.

This should be possible via both Ruby and env variables (`TEST_VERNIER_HOOKS=rails`).

## See also

- https://github.com/jhawthorn/vernier/blob/f8976625d7b5cfeeddb8471750f1ca93cf10d5aa/lib/vernier/collector.rb#L31
- https://github.com/jhawthorn/vernier/issues/64

PR #295, opened 2024-05-18
add hooks parameter to vernier

lHydra:

Add hooks configuration parameter for `TestProf::Vernier` to add event markers from Active Support Notifications

### Checklist

Closes https://github.com/test-prof/test-prof/issues/289

### DIFF
--- a/lib/test_prof/vernier.rb
+++ b/lib/test_prof/vernier.rb
@@ -19,7 +19,7 @@ module TestProf
   module Vernier
     # Vernier configuration
     class Configuration
-      attr_accessor :mode, :target, :interval
+      attr_accessor :mode, :target, :interval, :hooks
 
       def initialize
         @mode = ENV.fetch("TEST_VERNIER_MODE", :wall).to_sym
@@ -27,6 +27,7 @@ module TestProf
 
         sample_interval = ENV["TEST_VERNIER_INTERVAL"].to_i
         @interval = (sample_interval > 0) ? sample_interval : nil
+        @hooks = (ENV["TEST_VERNIER_HOOKS"] == "rails") ? :rails : nil
       end
 
       def boot?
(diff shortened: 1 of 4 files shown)


lib/test_prof/vernier.rb
```diff
@@ -19,14 +19,15 @@ module TestProf
...
-      attr_accessor :mode, :target, :interval
+      attr_accessor :mode, :target, :interval, :hooks
 
       def initialize
         @mode = ENV.fetch("TEST_VERNIER_MODE", :wall).to_sym
         @target = (ENV["TEST_VERNIER"] == "boot") ? :boot : :suite
 
         sample_interval = ENV["TEST_VERNIER_INTERVAL"].to_i
         @interval = (sample_interval > 0) ? sample_interval : nil
+        @hooks = (ENV["TEST_VERNIER_HOOKS"] == "rails") ? :rails : nil
```
palkan, line 30: We should probably take into account possible future named hooks; so, having smith like: `ENV["TEST_VERNIER_HOOKS"]&.split(",").map(&:to_sym)` would be better.

palkan (changes requested):
Thanks!

Let's try to add a test? At least a smoke test (simply checking that when the hook is provided we do not fail). We can also try to parse the report file and check if it contains some ActiveSupport event; though not sure if it's doable.

lib/test_prof/vernier.rb
lHydra: done

lHydra:
> Thanks!
> 
> Let's try to add a test? At least a smoke test (simply checking that when the hook is provided we do not fail). We can also try to parse the report file and check if it contains some ActiveSupport event; though not sure if it's doable.

Report contains `markerSchema`, which includes rails events when the hook is enabled. So in the test, I decided to check for a rails event

palkan (approved):
🚀

Issue #292, opened 2024-05-04
Track traits and overrides with FactoryProf


palkan:
## Context

Currently, we only track the factory name/id when collecting information about factories usage. However, for refactoring and context extraction purposes it would be useful to also see common _variations_ (traits and attribute overrides).

## TODO

Add _variations_ information to the FactorProf data and print it in the `FPROF=1` report. Example format:

```sh
[TEST PROF INFO] Factories usage

Total: 15285
...

    name           total   top-level   total time    time per call      top-level time           
    user            6091        2715    115.7671s          0.0426s            50.2517s 
     .admin          123         715     15.7671s          0.0466s             5.2517s 
     [name,role]      25          11       7.671s          0.0666s             1.2517s 
    post            2142        2098     93.3152s          0.0444s            92.1915s       
     .draft[tags]     12          12      9.3152s          0.164s             42.1915s        
  ...

```

In the example above, `.xxx` indicates a trait and `[a,b]` indicates the overrides keys, e.g., `create(:user, :admin)` is an `.admin` variation, while `create(:post, :draft, tags: ["a"])`—`.draft[tags]`. We do not distinguish values, only _schemas_.

## Implementation notes

- FactoryProf collector should receive the variation name from the outside and should not deal with traits, overrides, etc. Thus, we need to extends the `#track` methods with the `variation: ` keyword argument

- Overrides keys and traits must be sorted (since the order doesn't matter)

- It makes sense to add a specific variation id for variations with too many traits or overrides ("too many" must be configurable); e.g., `create(:post, title: "RailsConf", publish_at: 2.days.since, tags: ["a", "b"], author: user)` would be marked as `[...]`.

PR #296, opened 2024-05-20
Track traits and overrides with FactoryProf

lHydra:

Add factory variants (such as traits/overrides) to the output in simple mode. This information will also be in the resulting json file.

Add `variants_limit` config parameter for `FactoryProf` to use special ID (`[...]`) for factories with too many traits/overrides. The default value is 2.

1. Do we need default values for the variations limit?
2. Traits can still be long and accordingly the output gets messed up. I wanted to know if this is worth fixing, or is this behavior generally expected?

### Checklist

Closes https://github.com/test-prof/test-prof/issues/292

### DIFF
--- a/lib/test_prof/factory_prof/factory_bot_patch.rb
+++ b/lib/test_prof/factory_prof/factory_bot_patch.rb
@@ -5,7 +5,19 @@ module TestProf
     # Wrap #run method with FactoryProf tracking
     module FactoryBotPatch
       def run(strategy = @strategy)
-        FactoryBuilders::FactoryBot.track(strategy, @name) { super }
+        variation = ""
+
+        if @traits || @overrides
+          unless @traits.empty?
+            variation += @traits.sort.join(".").prepend(".")
+          end
+
+          unless @overrides.empty?
+            variation += @overrides.keys.sort.to_s.gsub(/[\\":]/, "")
+          end
+        end
+
+        FactoryBuilders::FactoryBot.track(strategy, @name, variation: variation.to_sym) { super }
       end
     end
   end
--- a/lib/test_prof/factory_prof/factory_builders/fabrication.rb
+++ b/lib/test_prof/factory_prof/factory_builders/fabrication.rb
@@ -15,8 +15,8 @@ module TestProf
           end
         end
 
-        def self.track(factory, &block)
-          FactoryProf.track(factory, &block)
+        def self.track(factory, **opts, &block)
+          FactoryProf.track(factory, **opts, &block)
         end
       end
     end
--- a/lib/test_prof/factory_prof/factory_builders/factory_bot.rb
+++ b/lib/test_prof/factory_prof/factory_builders/factory_bot.rb
@@ -18,9 +18,9 @@ module TestProf
             defined? TestProf::FactoryBot
         end
 
-        def self.track(strategy, factory, &block)
+        def self.track(strategy, factory, **opts, &block)
           return yield unless strategy.create?
-          FactoryProf.track(factory, &block)
+          FactoryProf.track(factory, **opts, &block)
         end
       end
     end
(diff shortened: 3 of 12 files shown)


lib/test_prof/factory_prof/printers/simple.rb
```diff
@@ -28,17 +28,54 @@ def dump(result, start_time:)
...
+            end
           end
 
           log :info, msgs.join("\n")
         end
+
+        private
+
+        def compress_variations(variations, variations_limit)
+          limited_variation = {
```
palkan, line 48: Let's rename it to `other_variation`

lib/test_prof/factory_prof/factory_bot_patch.rb
```diff
@@ -5,7 +5,19 @@ module FactoryProf
...
-        FactoryBuilders::FactoryBot.track(strategy, @name) { super }
+        variation = ''
+
+        if @traits || @overrides
+          unless @traits.empty?
+            variation += @traits.join('.').prepend('.')
+          end
+
+          unless @overrides.empty?
+            variation += @overrides.keys.to_s.gsub(/[\\":]/, '')
```
palkan, line 16: The keys must be sorted, we shouldn't distinguish similar key sets.

lib/test_prof/factory_prof/factory_builders/fabrication.rb
```diff
@@ -15,8 +15,8 @@ def self.patch
           end
         end
 
-        def self.track(factory, &block)
-          FactoryProf.track(factory, &block)
+        def self.track(factory, variation:, &block)
+          FactoryProf.track(factory, variation: variation, &block)
```
palkan, line 19: Let's use propagation here: `**opts`

lib/test_prof/factory_prof/factory_builders/factory_bot.rb
```diff
@@ -18,9 +18,9 @@ def self.patch
             defined? TestProf::FactoryBot
         end
 
-        def self.track(strategy, factory, &block)
+        def self.track(strategy, factory, variation:, &block)
           return yield unless strategy.create?
-          FactoryProf.track(factory, &block)
+          FactoryProf.track(factory, variation: variation, &block)
```
palkan, line 23: Same here

palkan (changes requested):
Thanks!

Let's move the other variation (`...`) check to the tracking method instead of dealing with this `compressed_variations` logic (and duplicating it).

We can use the `:__other__` variation name for off limit variations and track it like that; only when we use the simple printer, we need to treat this key differently (show at  the bottom of the variations list).

lHydra:
> Let's move the other variation (`...`) check to the tracking method instead of dealing with this `compressed_variations` logic (and duplicating it).

So we immediately write other variation (`...`) to the result if the limit is exceeded? I.e. if we export data to json, we also will see “compressed” data there?

lib/test_prof/factory_prof/factory_bot_patch.rb
lHydra: Done

lib/test_prof/factory_prof/factory_builders/fabrication.rb
lHydra: Done

lib/test_prof/factory_prof/factory_builders/factory_bot.rb
lHydra: Done

palkan:
>  I.e. if we export data to json, we also will see “compressed” data there?

Good point. I think, for now it's okay to have "__other__" in the JSON output as well.

lib/test_prof/factory_prof/printers/simple.rb
lHydra: I moved this logic to the #track method https://github.com/test-prof/test-prof/pull/296/files#diff-5bdf54a38f0e448125dcf0c2342a75e4c65f46ebb5a85f5ec8510dbba1b55f9eR139

palkan:
Thanks! Merged (via f1ee78b882935c0e4e3a5f77d66a30429fa42ec3)

PR #304, opened 2024-08-20
fix: dont load active record connection in dry-run

devinburnette:
Was getting `PG::ConnectionBad` and `ActiveRecord::ConnectionNotEstablished` errors when I added `let_it_be` and ran some `rswag` rake tasks in development which runs tests in `--dry-run` mode.. I would expect `test-prof` functionality to be fully compatible with `rspec --dry-run`.. So I think the right behavior here is to skip setting up an active record connection when it's a `--dry-run` because it seems the current behavior of `rspec --dry-run` doesn't have any assumptions about having a running database.

Moved the adapter require into the module so that it can be lazy loaded and protected behind a new conditional to determine if we're in rspec's dry-run mode

n/a

### Checklist

-->

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## master (unreleased)
 
+## 1.4.1 (2024-08-20)
+- Skips loading the ActiveRecord adapter for runs where RSpec --dry-run mode is enabled. ([@devinburnette][])
+
 ## 1.4.0 (2024-08-12)
 
 - AnyFixture: Disable fixtures cache within _clean fixture_ context. Automatically _refind_ records when using `#fixture`. ([@palkan][])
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## master (unreleased)
 
+## 1.4.1 (2024-08-20)
```
palkan, line 5: It's better to not touch versions/dates in PRs—you never know when it's gonna be merged 🙂

palkan:
Thanks! I'll publish a new release shortly

devinburnette:
Thanks @palkan for merging. Rough estimate on a new release?

palkan:
It's already out, v1.4.1 🙂

Issue #313, opened 2024-12-05
Unexpected RStamp  behavior with `\'` in a single quoted description


elasticspoon:
Hi! I have the following issue when using RStamp with:
- Ruby 3.3.6
- Rspec 3.13
- Rails 7.2.1

I have a example in a spec like this (please don't ask why this is a thing in the codebase)

```rb
it 'has a \' in the description' do # with comment
  expect(subject.body).to eq("Not Found")
end
```

I run `RSTAMP="todo" rspec that_spec.rb` and I expect:

```rb
it 'has a in the description', :todo do # with comment
  expect(subject.body).to eq("Not Found")
end
```

I get

```rb
it "has a  # with comment
  in the description", :todo do # with comment
  expect(subject.body).to eq("Not Found")
end
```

Which is not actually technically an invalid test but it seems pretty unexpected.

I don't think I have enough context to come up with a fix but I detailed what causes the issue here https://github.com/test-prof/test-prof/pull/314

palkan:
Closed by #314

PR #314, opened 2024-12-05
Fix/escaped single quote

elasticspoon:
Related to https://github.com/test-prof/test-prof/issues/313

This is a PR with some changes to the rstamp_tests to more closely reflect reality and a test case to illustrate the behavior in this issue.

### Test Changes

When RStamp reads in files it uses `File.readlines` which results in array of strings ending in newlines:

```rb
# somefile.rb
puts 'foo'
puts 'bar'

# somewhere else
File.readlines('somefile.rb')
=> ["puts 'foo'\n", "puts 'bar'\n"]
```

Meanwhile in the tests we take a heredoc and split it along the newline:

```rb
let(:source) do
  <<~CODE
    it "doesn't do what it should do" do
      expect(subject.body).to eq("OK")
    end
  CODE
end

let(:code) { source.split("\n") }

code
=> ["it \"doesn't do what it should do\" do", "expect(subject.body).to eq(\"OK\")", "end"]
```

This means that the code inputs do not have trailing newlines. It does not matter with the current test cases BUT with those changes the following added test case fails.

```rb
context "with escaped single quotes" do
  let(:source) do
    <<~code
      it 'has a \\' in the description' do # with comment
        expect(subject.body).to eq("not found")
      end
    code
  end

  let(:expected) do
    <<~code
      it "has a in the description", :todo do # with comment
        expect(subject.body).to eq("not found")
      end
    code
  end

  specify do
    is_expected.to eq 0
    expect(code.join).to eq expected
  end
end
```

### Failing Test Case

The reason the test case fails is the following code

```rb
# lib/test_prof/rspec_stamp.rb
def stamp_example(example, tags)
  # ...
  replacement = "\\1#{parsed.fname}#{need_parens ? "(" : " "}" \
                "#{[desc, tags_str, htags_str].compact.join(", ")}" \
                "#{need_parens ? ") " : " "}\\3"
  # replacement => "\\1it \"has a \\' in the description\", :todo \\3"

  # EXAMPLE_RXP = /(\s*)(\w+\s*(?:.*)\s*)(do|{)/.freeze
  example.sub!(EXAMPLE_RXP, replacement)
```

In this case the `\\1` and `\\3` will get replacement with their respective match groups, but the `\'` got escaped in such a way that is has turned into a match group. `$'` is a global variable that holds the match group for the string after the match [(see here)](https://docs.ruby-lang.org/en/2.2.0/Regexp.html#class-Regexp-label-Special+global+variables). So anything after `do` or `{` could be replaced there, except only in this case is it valid syntax.

### Fixing

I am not sure what the best fix is but I feel it would be something revolving around not modifying the strings in the `stamp_example` function and instead building up a replacement string and passing it back. However, I am unsure if this edge case merits the rework and if there was some other reason that the string were modified in place.

I do find the modification of strings in place to be pretty surprising.

### DIFF
--- a/lib/test_prof/rspec_stamp.rb
+++ b/lib/test_prof/rspec_stamp.rb
@@ -153,14 +153,16 @@ module TestProf
           end
         end
 
-        replacement = "\\1#{parsed.fname}#{need_parens ? "(" : " "}" \
-                      "#{[desc, tags_str, htags_str].compact.join(", ")}" \
-                      "#{need_parens ? ") " : " "}\\3"
+        replacement = proc do
+          $1 + "#{parsed.fname}#{need_parens ? "(" : " "}" \
+          "#{[desc, tags_str, htags_str].compact.join(", ")}" \
+          "#{need_parens ? ") " : " "}" + $3
+        end
 
         if config.dry_run?
-          log :info, "Patched: #{example.sub(EXAMPLE_RXP, replacement)}"
+          log :info, "Patched: #{example.sub(EXAMPLE_RXP, &replacement)}"
         else
-          example.sub!(EXAMPLE_RXP, replacement)
+          example.sub!(EXAMPLE_RXP, &replacement)
         end
         true
       end
--- a/spec/test_prof/rspec_stamp_spec.rb
+++ b/spec/test_prof/rspec_stamp_spec.rb
@@ -150,6 +150,29 @@ describe TestProf::RSpecStamp do
       end
     end
 
+    context "with escaped single quotes" do
+      let(:source) do
+        <<~CODE
+          it 'has a \\' in the description' do # with comment
+            expect(subject.body).to eq("Not Found")
+          end
+        CODE
+      end
+
+      let(:expected) do
+        <<~CODE
+          it "has a in the description", :todo do # with comment
+            expect(subject.body).to eq("Not Found")
+          end
+        CODE
+      end
+
+      specify do
+        is_expected.to eq 0
+        expect(code.join).to eq expected
+      end
+    end
+
     context "one-liner" do
       let(:source) do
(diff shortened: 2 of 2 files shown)


palkan (changes requested):
Thanks for the fix!

Could you please take a look at the failing spec?

elasticspoon:
> Thanks for the fix!
> 
> Could you please take a look at the failing spec?

Sorry for the confusion. I did not actually implement the fix just made a failing spec to demonstrate the issue. I'm down to do it tho. 

Only question I had is: the code mutates an array of strings when it does the replacements. Should I keep it that way or can I return new strings instead?

palkan:
> I did not actually implement the fix just made a failing spec to demonstrate the issue. I'm down to do it tho.

Ah, got it.

> the code mutates an array of strings when it does the replacements. Should I keep it that way or can I return new strings instead?

I don't remember why it has been implemented this way (I believe, this code stayed as is since the very first version back from 2017). Creating new strings would require quite a refactoring.

I wonder if instead we can: 1) get rid of `\\<n>` substitutes and use, for example,`example.gsub!(regexp) do ... end` to replace parts; 2) add some escaping, so we don't accidentally introduce  replacements.

elasticspoon:
> 1) get rid of `\\<n>` substitutes and use, for example,`example.gsub!(regexp) do ... end` to replace parts

That approach works nicely.

> 2) add some escaping, so we don't accidentally introduce replacements.

Not 100% sure what you would like to escape. Or was that meant as a alternative option to 1?

palkan:
> Not 100% sure what you would like to escape. Or was that meant as a alternative option to 1?

Yeah, that was an alternative idea. Nvm, the current implementation looks good 👍 Thanks!

PR #315, opened 2025-01-08
Add YJIT detection

yaroslav:
Ads simple YJIT detection and warnings.

It would be great if we could disable YJIT when enabled, but it is currently not possible.

Reasoning:

On dev machines, a developer might have set up YJIT by default, because it is cool and someone recommended it. On production, people might test in a production environment, the same Rails runs in, or using a third party pre-setup container that launches Ruby with YJIT because it is "faster". In reality, they are losing %% of their test suite performance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+- Add a warning to check if YJIT is enabled, as it is known to slow down test environments. ([@yaroslav][])
+
 ## 1.4.3 (2024-12-18)
 
 - Fix handling new (lazy) connection pools in `before_all`. ([@palkan][])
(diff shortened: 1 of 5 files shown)


palkan:
Thanks!

I don't think we need a dedicated tool for that, just mentioning this in the [Playbook](https://test-prof.evilmartians.io/playbook) would be enough.

There is a plan (okay, an idea) of a TestProf Audit tool that would perform various checks like that against the test env. But we're not there yet.

yaroslav:
OK with not merging, no problem.

Not sure I'm on the same page re: documentation. Not a lot of people are willingly reading documentation, and probably nobody will _re-read_ the Playbook after learning the toolkit the first time.

I'm pretty sure that if we can detect a rule, be that YJIT usage, or excessive logging, or password rehashing, we should warn and potentially optimize it ourselves. 

The YJIT problem is not made up, of course https://github.com/evilmartians/test-prof-workshop/blob/52906a1b515b55e32cd9deb88c6428690859e0c7/config/initializers/enable_yjit.rb

Is the Audit thing (or, rather, an idea) supposed to be a part of TestProf or a separate Gem?..

palkan:
> Is the Audit thing (or, rather, an idea) supposed to be a part of TestProf or a separate Gem?..

Ideally, separate gem; but first we need to extract common utils (like RSpec/Spring/whatever detection, logging, etc.)

PR #319, opened 2025-03-15
Mention vernier.prof, a tenderlove/profiler host

ellnix:
This seems to have been added last year in vernier's readme: https://github.com/jhawthorn/vernier/commit/70d463c860fc7d6ff1f3dec392a6c984745c4254

It's an instance of the tenderlove/profiler firefox profiler fork, which means it's much more suitable for ruby. In my limited testing, only the fork was able to show source.

Apologies if I misunderstood or missed something, this is my first time using both this gem and vernier.


palkan:
Thanks!

> Apologies if I misunderstood or missed something, this is my first time using both this gem and vernier.

Welcome to the family of TestProf users and contributors!

PR #346, opened 2026-07-08
Add logger as a runtime dependency

moznion:

`lib/test_prof/core.rb` requires "logger", which is no longer a default gem as of Ruby 4.0.
ref: https://www.ruby-lang.org/en/news/2025/12/18/ruby-4-0-0-preview3-released/#stdlib-updates

This might become the issue of `LoadError: cannot load such file -- logger` so it'd be better to fix this problem.

Similar issues:

- https://github.com/phusion/passenger/issues/2642
- https://github.com/aws/aws-sdk-ruby/issues/3206
- https://github.com/sidekiq/sidekiq/issues/6319

Declare it explicitly so test-prof does not rely on it being pulled in transitively (e.g. via activesupport).

n/a

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+- Add `logger` as a runtime dependency (it's no longer a default gem in Ruby 4.0). ([@moznion][])
+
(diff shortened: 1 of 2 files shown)


palkan:
Thanks!

PR #348, opened 2026-07-09
Fix RSpecDissect crash when let is evaluated on a non-example thread

### LINKED ISSUE #347: RSpecDissect crashes with `undefined method 'last' for nil` when a `let` is evaluated on a non-example thread
### Summary

Since the RSpecDissect "let breakdowns" feature added in **1.6.0**, `RSpecDissect.current_span` and `track` dereference the thread-local `Thread.current[:_rspec_dissect_spans_stack]` unconditionally. That thread-local is only initialized (via `reset!`) on the RSpec **example** thread.

If a memoized helper (`let`) is evaluated on a **different** thread — the common real-world case being a Capybara `:js`/system spec, where the application runs in a separate Puma server thread — RSpecDissect raises:

```
NoMethodError: undefined method 'last' for nil
  test_prof/rspec_dissect.rb:89:in 'current_span'
  test_prof/rspec_dissect.rb:97:in 'track'
  test_prof/rspec_dissect.rb:29:in 'MemoizedInstrumentation#fetch_or_store'
```

In a Rails/Capybara app this surfaces as an `ActionView::Template::Error` mid-render and a `502` in the browser console, so the backtrace points at unrelated application code and the true cause is very hard to locate.

### Regression bisect

Confirmed via A/B with an identical spec and `RD_PROF=1`:

| test-prof | Result |
|-----------|--------|
| 1.5.2     | RSpecDissect enabled, **passes** |
| 1.6.1     | RSpecDissect enabled, **`nil.last` crash** |

The thread-local span-stack code (`MemoizedInstrumentation`, `Span`, `span_stack`, `current_span`) was introduced in 1.6.0 ("Upgrade RSpecDissect to show total setup time and let breakdowns"); it does not exist in 1.5.2, which is why 1.5.2 is unaffected.

### Minimal reproduction (standalone, no Rails/Capybara needed)

`Gemfile`:
```ruby
source "https://rubygems.org"
gem "rspec", "~> 3.13"
gem "test-prof", "= 1.6.1"
gem "activesupport"
```

`spec_helper.rb`:
```ruby
require "active_support"
require "active_support/notifications"
require "test_prof"
```

`repro_spec.rb`:
```ruby
require_relative "spec_helper"

RSpec.describe "RSpecDissect cross-thread let" do
  let(:value) { 42 }

  it "reads a let from another thread" do
    result = nil
    Thread.new { result = value }.join
    expect(result).to eq(42)
  end
end
```

Run:
```
RD_PROF=1 bundle exec rspec repro_spec.rb
```

Result on test-prof 1.6.1 (Ruby 3.4.9):
```
[TEST PROF INFO] RSpecDissect enabled
#<Thread:0x... repro_spec.rb:8 run> terminated with exception (report_on_exception is true):
test_prof/rspec_dissect.rb:89:in 'TestProf::RSpecDissect.current_span': undefined method 'last' for nil (NoMethodError)

        Thread.current[:_rspec_dissect_spans_stack].last
                                                   ^^^^^
	from test_prof/rspec_dissect.rb:97:in 'TestProf::RSpecDissect.track'
	from test_prof/rspec_dissect.rb:29:in 'TestProf::RSpecDissect::MemoizedInstrumentation#fetch_or_store'
1 example, 1 failure
```

Without `RD_PROF` (or on test-prof 1.5.2) the same spec passes.

### Root cause

`MemoizedInstrumentation#fetch_or_store` wraps every `let` in `RSpecDissect.track`, and both `track` and `current_span` assume the span-stack thread-local is initialized:

```ruby
def current_span
  Thread.current[:_rspec_dissect_spans_stack].last   # nil.last when unset
end

def span_stack
  Thread.current[:_rspec_dissect_spans_stack]         # nil on a non-example thread
end

def track(type, id: nextid, **meta)
  span = Span.new(id, current_span&.id, type, 0.0, meta)
  span_stack << span                                  # nil << ... also raises
  ...
ensure
  span_stack.pop                                      # nil.pop also raises
end
```

`reset!` only initializes it on the example thread:

```ruby
def reset!
  @last_id = 1
  @spans = []
  Thread.current[:_rspec_dissect_spans_stack] = []
end
```

### Proposed fix

Lazily initialize the span stack so access from a non-example thread is a safe no-op (spans created off the example thread are simply not part of that example's setup timing):

```ruby
def span_stack
  Thread.current[:_rspec_dissect_spans_stack] ||= []
end

def current_span
  span_stack.last
end
```

Verified: with this change applied to 1.6.1, the reproduction above passes with `RD_PROF=1` while RSpecDissect stays enabled. This fixes `current_span` and the `<<`/`pop` in `track` in one place.

dstull:
RSpecDissect stores its span stack in a thread-local (Thread.current[:_rspec_dissect_spans_stack]) that is only initialized via reset! on the RSpec example thread. When a memoized helper (let) is evaluated on a different thread -- most commonly a Capybara :js/system spec, where the app runs in a separate Puma server thread -- current_span and track dereference the uninitialized thread-local and raise NoMethodError: undefined method 'last' for nil.

Lazily initialize the span stack in span_stack so access from a non-example thread is a safe no-op (spans created off the example thread are simply not collected as part of that example's setup timing). This fixes both current_span and the << / pop in track.

Introduced in 1.6.0 with the RSpecDissect let-breakdowns feature; 1.5.2 is unaffected.

Closes #347

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+- Fix `RSpecDissect` crashing with `undefined method 'last' for nil` when a `let` is evaluated on a non-example thread (e.g. a Capybara `:js`/system spec server thread). The span-stack thread-local is now lazily initialized so off-example-thread access is a safe no-op.
+
 ## 1.6.1 (2026-04-02)
 
 - Require MFA to publish the gem.
--- a/lib/test_prof/rspec_dissect.rb
+++ b/lib/test_prof/rspec_dissect.rb
@@ -86,11 +86,11 @@ module TestProf
       end
 
       def current_span
-        Thread.current[:_rspec_dissect_spans_stack].last
+        span_stack.last
       end
 
       def span_stack
-        Thread.current[:_rspec_dissect_spans_stack]
+        Thread.current[:_rspec_dissect_spans_stack] ||= []
       end
 
       def track(type, id: nextid, **meta)
--- a/spec/bugs/fixtures/rspec_dissect_cross_thread_fixture.rb
+++ b/spec/bugs/fixtures/rspec_dissect_cross_thread_fixture.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+$LOAD_PATH.unshift File.expand_path("../../../../../lib", __FILE__)
+require "test-prof"
+
+# Evaluating a memoized helper (`let`) from a thread other than the RSpec
+# example thread must not crash RSpecDissect. The span-stack thread-local is
+# only initialized on the example thread; reading a `let` elsewhere (e.g. a
+# Capybara server thread) previously raised `NoMethodError: undefined method
+# 'last' for nil`.
+describe "Something" do
+  let(:value) { 42 }
+
+  it "reads a let from another thread" do
+    result = nil
+    Thread.new { result = value }.join
+    expect(result).to eq 42
+  end
+end
--- a/spec/bugs/rspec_dissect_cross_thread_spec.rb
+++ b/spec/bugs/rspec_dissect_cross_thread_spec.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+describe "RSpecDissect with cross-thread let", type: :integration do
+  specify do
+    output = run_rspec(
+      "rspec_dissect_cross_thread",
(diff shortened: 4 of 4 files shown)


palkan:
Awesome, thanks!

PR #351, opened 2026-07-21
Fix before_all with parallel Minitest 6

GabrielNagy:

Minitest 6 defines `run_suite` as a [class method](https://github.com/minitest/minitest/blob/v6.0.6/lib/minitest.rb#L452-L486), but the existing check looked for an instance method. `TestProf` therefore installed the Minitest 5 `run_one_method` hook, which made Rails [call a method](https://github.com/rails/rails/blob/v8.0.5/activesupport/lib/active_support/testing/parallelization/worker.rb#L49-L54
) that Minitest 6 no longer provides.

Detect Minitest 6 through its class API and hook the per-test run method that Rails 8 workers use. Keep the existing `run_one_method` path for Minitest 5.

Here's a repro script which fails on master and passes with this PR:
```ruby
require "minitest"
require "minitest/test"
require "active_support/core_ext/module/attribute_accessors"
require "test_prof/recipes/minitest/before_all"

TestProf::BeforeAll.adapter = TestProf::BeforeAll::NoopAdapter

class ExampleTest < Minitest::Test
  include TestProf::BeforeAll::Minitest

  before_all { @value = 1 }

  def test_value
    assert_equal 1, @value
  end
end

# This is the same API choice that a Rails parallel worker makes.
result = if Minitest.respond_to?(:run_one_method)
  Minitest.run_one_method(ExampleTest, :test_value)
else
  ExampleTest.new(:test_value).run
end

puts "Manual run passed: #{result.passed?}"
```

```sh-session
> BUNDLE_GEMFILE=gemfiles/activerecord8.gemfile bundle exec ruby -Ilib repro.rb
~/code/test-prof/lib/test_prof/recipes/minitest/before_all.rb:23:in 'run_one_method': super: no superclass method 'run_one_method' for module Minitest (NoMethodError)
        from repro.rb:20:in '<main>'
```

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+- Fix `before_all` with parallel Minitest 6. ([@GabrielNagy][])
+
 ## 1.6.3 (2026-07-20)
 
 - Fix `before_all` firing N times in Minitest 6. ([@palkan][])
--- a/lib/test_prof/recipes/minitest/before_all.rb
+++ b/lib/test_prof/recipes/minitest/before_all.rb
@@ -3,22 +3,13 @@
 require "test_prof/before_all"
 
 Minitest.singleton_class.prepend(Module.new do
-  attr_reader :previous_klass
+  attr_accessor :previous_klass
   @previous_klass = nil
+end)
 
-  # Minitest 6+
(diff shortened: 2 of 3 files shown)


palkan:
Thanks!

PR #353, opened 2026-08-01
add ability to track callbacks runs

lHydra:
Closes https://github.com/test-prof/test-prof/discussions/350.

Implement the model-callback profiling requested in #350 — as an EventProf plugin

It adds a new `callback.run` event so you can see how much time is spent in
`ActiveSupport` callbacks (`before`/`after`) across a test run, per callback:

```sh
EVENT_PROF=callback.run bundle exec rspec
```

Add new **callback.run** custom event (`lib/test_prof/event_prof/custom_events/callback_run.rb`).
It prepends `ActiveSupport::Callbacks::CallTemplate.singleton_class` and wraps each compiled
callback lambda in a lightweight `CallbackInvocation` proxy that carries `kind/name/filter`,
then instruments its #call via `EventProf.monitor`

The `CallTemplate` monkey-patch relies on `ActiveSupport::Callbacks internals` (`CallTemplate.build +
#make_lambda, Callback#kind/#name/@filter`). I've verified the approach is stable across
Rails 5.1 → 8.x / main.  But as I understand it, we only need support for version 6 and higher? So should I add the guard to check Rails version?

### Checklist


lHydra:
Added `CallbackInvocation#to_proc` since Rails 8.2 block-converts the `make_lambda` result https://github.com/rails/rails/blob/main/activesupport/lib/active_support/callbacks.rb#L289, which caused `TypeError` under railsmaster.gemfile.

